PR #14, opened 2020-05-25
Added bindata tool and reorganized tests

joshuahannan:
* Reorganized the tests similar to how the flow-ft repo looks
* added bindata tool for the `contracts` module
* add makefile
* consolidate cadence code to `src`


src/contracts/Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: test
```
psiemens, line 1: This file should actually be in `/contracts` instead of `/src/contracts`

PR #29, opened 2020-11-05
Update mint_nft.cdc

iwill610:
I was having some issues with this transaction and I brought it up in the discord.   The solution I got from the discord was to put the address inside getAccount.  I later tried the change I made below and that worked as well.

let recipient = getAccount(signer.address) in line 22

### DIFF
--- a/transactions/mint_nft.cdc
+++ b/transactions/mint_nft.cdc
@@ -19,7 +19,7 @@ transaction(recipient: Address) {
 
     execute {
         // Get the public account object for the recipient
-        let recipient = getAccount(recipient)
+        let recipient = getAccount(signer.address)
 
         // Borrow the recipient's public NFT collection reference
         let receiver = recipient


joshuahannan:
@iwill610 Sorry for not responding to this sooner! This is a good idea and would definitely make it easier for playground and VSCode users, but when we're using it with the SDK, we want to use transaction arguments because that is the proper way to do it. Sorry for the confusion. I should add some comments to clarify that.

iwill610:
@joshuahannan All good.  I just realized that the code I have for this transaction is a little different so my solution wouldn't have worked anyway.  Could you give me an example of how you pass arguments through the transactions?  I think I read that it should similar to functions, but transactions don't have names, so that left me a little stuck.

joshuahannan:
Passing arguments with the playground or VSCode isn't possible yet, but you can do it with JS or Go. Here is an example of creating a transaction in Go and adding arguments to it. You'll probably need to ask in the javascript channel in discord if JS is the language you're using though

iwill610:
Thank you

PR #42, opened 2021-04-29
fixed mixed tabs and spaces in example contracts

solipsis:
Mixing of tabs and spaces makes the standard look bad in certain editors like vim


joshuahannan (approved):
Thank you!

Issue #10, opened 2020-05-08
CONTRACT: Create NFT Forwarding Contract


joshuahannan:
### Issue To Be Solved
Some users may want to forward the NFTs that are sent to them to a different account's collection. Make a contract that defines a resource type that implements the `Receiver` interface from `NonFungibleToken` and deposits the NFT into another account. 

This is similar to the `TokenForwarding` contract from the flow-ft repo

### (Optional): Suggest A Solution
* Write the contract
* Test it with the `transfer_nft.cdc` transaction

phouse512:
I know this is old but I'd like to take a stab at it if it's still relevant. @joshuahannan do you think this would still be helpful to have?

joshuahannan:
@phouse512 we don't have a critical need for it, but it is always helpful to have more good smart contract examples of different use cases! It would be awesome if you could take a crack at it. I can help answer questions if needed! :)

phouse512:
thanks @joshuahannan, I just published a PR for this - let me know if that's what you were thinking or if I missed what you were aiming for

PR #43, opened 2021-05-15
add example NFT forwarder, with tests

phouse512:
adds an `NFTForwarding` contract that allows for an owner to redirect NFTs to another recipient. closes issue #10 

- adds `create_forwarder.cdc` transaction for creating forwarder object with recipient as argument
- adds `TokenForwarding.cdc` contract that holds `NFTForwarding` contract, which implements the `Receiver` interface
- adds `lib/go/test/forwarding_test.go` and utilities to help test forwarder in action


phouse512:
I'm not sure why the builds haven't been running - let me know if there is anything I need to do to fix that

contracts/TokenForwarding.cdc
```diff
@@ -0,0 +1,49 @@
...
+
+*/
+
+import NonFungibleToken from 0x02
+
+pub contract NFTForwarding {
+
+    // Event that is emitted when NFTs are forwarded to the target receiver
+    pub event ForwardedToken(id: UInt64, from: Address?)
+
```
joshuahannan, line 13: Maybe it would be good to include a canonical NFT forwarder path that could be used when storing the forwarder? something like this
```suggestion
let nftForwarderPath: StoragePath
```

contracts/TokenForwarding.cdc
```diff
@@ -0,0 +1,49 @@
...
+            let tokenId = token.id
+            receiverRef.deposit(token: <- token)
+
+            emit ForwardedToken(id: tokenId, from: self.owner?.address)
+        }
+
+        // changeRecipient allows the forwarding target to be updated
+        pub fun changeRecipient(_ newRecipient: Capability) {
+            pre {
+                newRecipient.borrow<&{NonFungibleToken.Receiver}>() != nil: "Could not borrow Receiver interface from the Capability"
```
joshuahannan, line 31:
```suggestion
                newRecipient.borrow<&{NonFungibleToken.Receiver}>() != nil: "Could not borrow NFT Receiver interface from the Capability"
```

contracts/TokenForwarding.cdc
```diff
@@ -0,0 +1,49 @@
...
+            }
+            self.recipient = newRecipient
+        }
+
+        init(recipient: Capability) {
+            pre {
+                recipient.borrow<&{NonFungibleToken.Receiver}>() != nil: "Could not borrow Receiver reference from the Capability"
```
joshuahannan, line 38:
```suggestion
                recipient.borrow<&{NonFungibleToken.Receiver}>() != nil: "Could not borrow NFT Receiver reference from the Capability"
```

transactions/create_forwarder.cdc
```diff
@@ -0,0 +1,23 @@
+import NonFungibleToken from 0xNFTADDRESS
+import ExampleNFT from 0xNFTCONTRACTADDRESS
+import NFTForwarding from 0xNFTFORWARDERADDRESS
+
```
joshuahannan, line 4: Can you add some comments to this transaction describing what contract it uses and how it works? Thanks!

joshuahannan (commented):
Looks really good! I left a few comments.

I'm not sure why the builds don't work when coming from a fork. I need to learn more about github actions so I can figure out what is going on here. If you have any ideas, I'm all ears.

contracts/TokenForwarding.cdc
phouse512: could you share why this makes sense? in this case it just gets defined in the creation transaction, and never really gets used again in the actual contract. Is this just to expose the path to future references of the forwarder?

transactions/create_forwarder.cdc
phouse512: added, see below!

contracts/TokenForwarding.cdc
joshuahannan: Yeah, whenever someone wants to add a forwarder to their account, they can use the contract-defined path instead of having to hard code it themselves

satyamakgec:
Closing this PR in the favour of #116

PR #52, opened 2021-12-20
Add NFT metadata standard

psiemens:
This RP integrates the new NFT metadata standard from [FLIP-0636](https://github.com/onflow/flow/blob/master/flips/20210916-nft-metadata.md) into this repository.


psiemens:
@bjartek for your consideration!

bjartek:
Excellent start! Until https://github.com/onflow/cadence/pull/1076 is merged this is the best we can do. 

When that is merged we can add the default implemented methods to the NFT spec. 

This will allow all new contracts to do like have done in example and they do not have to do anything when the standard is introduced in the future it will just work for them.

contracts/ExampleNFT.cdc
```diff
@@ -12,18 +13,44 @@ pub contract ExampleNFT: NonFungibleToken {
     pub event Withdraw(id: UInt64, from: Address?)
     pub event Deposit(id: UInt64, to: Address?)
 
-    pub resource NFT: NonFungibleToken.INFT {
+    pub resource NFT: NonFungibleToken.INFT, Views.Resolver {
         pub let id: UInt64
 
-        pub var metadata: {String: String}
+        pub let name: String
```
joshuahannan, line 19: Would it be good to include some comments about what each field and function does?

contracts/Views.cdc
```diff
@@ -0,0 +1,28 @@
+pub contract Views {
```
joshuahannan, line 1: Like my other comment, can we include a bunch of comments in the code here to show what each part is for?

joshuahannan (approved):
Mostly looks good! Basically just requesting some more comments. Also, are you planning on adding tests for the scripts for getting the views?

bjartek (changes requested):
I tried to run the tests locally but they fail.

README.md
```diff
@@ -115,6 +115,57 @@ let collection = account.getCapability(/public/ExampleNFTCollection)
...
+such as an on-chain creator biography or an off-chain video clip.
+
+### Display view
+
+The [`Display`](contracts/Metadata.cdc#L8-L22) view is the most basic metadata view.
+It returns the minimum information required to render an NFT in most applications.
+
+```swift
+import ExampleNFT from "..."
+import Metadata from "..."
```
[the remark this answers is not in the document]
psiemens: Good flag. So you're saying that contracts named `Metadata` already exist and are in use?

I'd like to avoid using `Views` because it's too generic. How about `MetadataViews`?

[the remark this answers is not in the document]
bjartek: I think that conctracts have the same namespace as structs and resources and I know that plenty have structs name Metadata. 

MetadataView will work.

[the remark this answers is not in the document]
psiemens: Yes you're right, structs and resource declarations will override the `Metadata` import.

bjartek:
What about adding 

Identity:
 - id
 - type
 - uuid
 
 This would fix a long standing problem with global identification. Since UUIDs are unique it woube be possible to create some powerful feature if this view can be almost mandatary. It is trivial to write aswell.

contracts/MetadataViews.cdc
```diff
@@ -0,0 +1,59 @@
...
+            self.description=description
+        }
+    }
+
+    // Thumbnail returns a thumbnail image for an object.
+    //
+    // Many NFT resources implement this view to provide 
+    // a simple visual representation of the NFT.
+    //
+    pub struct Thumbnail {
```
bjartek, line 47: Personally i do not like that this and Display is not split in two. Both for .find and blocto this would be that we have to do almost 2x the operations and grow complexity. I would rather the Display struct was even thicker. 

Fields that find needs to display sometehing "propperly" are
 - title
 - thumbnailUrl
 - sourceLink (in many cases this can be derrived from the id, so either we need id here or identity)
 - source (but we can get that from the collection)
 - tags (for filtereing)
 
 Also the Alchemy api has the following fields
 `pub let contract: NFTContract
    pub let id: UInt64
    pub let uuid: UInt64?
    pub let title: String?
    pub let description: String?
    pub let external_domain_view_url: String?
    pub let token_uri: String?
    pub let media: [NFTMedia?]
    pub let metadata: {String: AnyStruct}`

from https://github.com/alchemyplatform/alchemy-flow-contracts/blob/main/src/cadence/scripts/getNFTs.cdc

contracts/ExampleNFT.cdc
```diff
@@ -12,18 +13,67 @@ pub contract ExampleNFT: NonFungibleToken {
...
+            thumbnail: String,
+        ) {
+            self.id = id
+            self.name = name
+            self.thumbnail = thumbnail
+            self.description = description
+        }
+    
+        pub fun getViews(): [Type] {
+            return [
```
bjartek, line 40: I think this should add String that just resolves to self.name
I also think it should add Idenitty.

bjartek:
Why was the ResolverCollection removed? Before default impl are here and we can say that NFT supports this out of the box it is impossible to know if a resource supports this standard if that is gone.

contracts/MetadataViews.cdc
psiemens: Yeah, the `Thumbnail` struct is trading convenience for flexibility. I think we have two goals:

- Put all the basic info in one place (reduce complexity)
- Support multiple image formats and storage locations (flexibility)

IMO a simple string field is too limiting. For example, I think a web image (i.e. an HTTP URI) should be represented differently from an IPFS image (i.e. a CID). Some people might want to do on-chain SVGs. In the future, maybe some NFTs will use something entirely new.

It's trivial to implement all of these options as individual views. But as you pointed out, that pushes complexity onto the caller; they have to figure out which thumbnail views an NFT implements. 

I wondered if we could solve this with a `thumbnail` field on `Display` that points to a new interface.

```swift
pub struct Display {
   pub let thumbnail: Thumbnail
}

pub struct interface Thumbnail {
  // Problem: what if the image cannot (or should not) be represented with a URI?
  fun uri(): String
}
```

But I don't think that works. We're really just back to the string problem -- what common interface can represent all thumbnail types? There isn't one, really.

In the end I think the flexibility of multiple structs is worth the extra work needed to build a complete NFT representation. When you say 2x the operations, are you worried about transaction gas fees or just general code complexity? I was thinking it'd be fairly easy to add a helper function that resolves the most common thumbnail views and returns all the data as a single struct.

psiemens:
> Why was the ResolverCollection removed? Before default impl are here and we can say that NFT supports this out of the box it is impossible to know if a resource supports this standard if that is gone.

My bad, added that back.

contracts/MetadataViews.cdc
psiemens: I'm 👍 for `sourceLink` but I think `source` should always come from `nft.getType()` to avoid counterfeits.

contracts/ExampleNFT.cdc
joshuahannan: What is Identity? Also, is there a specific definition for what name would mean across different NFT projects or is it completely up to the developer?

contracts/MetadataViews.cdc
bjartek: But the type of the nft does not have enough information to distinguish sources when using a shared minter. Take the example with counterfeit NFT socks from flowverse. If somebody else mints a sock on Rarible there is no way of distinguish them since they both used the same shared minter.

bjartek: > But I don't think that works. We're really just back to the string problem -- what common interface can represent all thumbnail types? There isn't one, really.

I do not see a better solution then a string.

contracts/ExampleNFT.cdc
bjartek: In my eyes name is up to the developer. 

My Identity suggestion from up above is because I think exposing UUID is something that should really be done on all resources. It is so much more convenient to work with a UUID compared to the type and the ID to uniquely identify a resource. The reason i added Type and ID there was in case somebody needs to do a mapping.

bjartek:
Another thing I think would be nice to have in the NFT would be some struct that act as pointers to an NFT in a collection. 

In storefront right now there is a Capability to withdraw en NFT in the contract itself and then there is the ID as an adjacent field. I feel it would be a lot better if this was an struct that resided in the Metadata resource (or another one) that allowed solutions to create

- a ReadPointer to an NFT, or really to a ViewResolver
- an AuthPointer that allows you to withdraw this one NFT

contracts/ExampleNFT.cdc
```diff
@@ -12,18 +13,67 @@ pub contract ExampleNFT: NonFungibleToken {
...
+    }
 
-        init(initID: UInt64) {
-            self.id = initID
-            self.metadata = {}
+    pub resource interface ExampleNFTCollectionPublic {
+        pub fun deposit(token: @NonFungibleToken.NFT)
+        pub fun getIDs(): [UInt64]
+        pub fun borrowNFT(id: UInt64): &NonFungibleToken.NFT
+        pub fun borrowExampleNFT(id: UInt64): &ExampleNFT.NFT? {
```
bjartek, line 68: borrowViewReolver needs to be here if we want to use that. If not remove it.

PR #55, opened 2022-01-05
Add Display, HTTPFile and IPFSFile metadata views 

psiemens:
This PR proposes three metadata views that together can be used to return the minimal representation of an NFT.

The code here is still under discussion -- this PR is a follow-up [from a previous conversation](https://github.com/onflow/flow-nft/pull/52#discussion_r774409201) with @bjartek.


bjartek:
As previously stated I think solutions should need to implement a single View to be displayable in blocto app or on .find. In my eyes the Display view is one alternative to that, the other would be the ERC721JsonUrl that we also have talked about previously @psiemens.

The following fields are what we need/want in .find
 - name
 - sourceURL: (the url to where this item can be viewed)
 - thumbnailURL: (the url to the thumbnail, we just autodetect if it starts with ipfs:/// and then it is ipfs)
 - mimeType: the type of the thumnail
 - source: The source of this Item, Note that a Type is not enough here, some contracts like Rarible/Versus aso can have many sources that share the same contract)

For .find we also have listPrice and listToken, but that is not relevant in this case. 

I cannot speak for blocto or the alchemy api but it looks to like they are collecting some of the same information as we do. 

The reason I want this as a single View is because it is easier to implement, and alot easier to fetch and display. If everybody have to collect multiple Views and join them together to a data structure afterwards it will just lead to yet more diversity.

contracts/MetadataViews.cdc
```diff
@@ -29,4 +29,55 @@ pub contract MetadataViews {
...
+    // Display is a basic view that includes the name and description
+    // of an object. Most objects should implement this view.
+    //
+    pub struct Display {
+        pub let name: String
+        pub let description: String
+
+        init(
+            name: String,
+            description: String,
```
turbolent, line 42:
```suggestion
            description: String
```

contracts/ExampleNFT.cdc
```diff
@@ -59,17 +38,22 @@ pub contract ExampleNFT: NonFungibleToken {
...
+                case Type<MetadataViews.Display>():
+                    return MetadataViews.Display(
                         name: self.name,
                         description: self.description,
-                        thumbnail: self.thumbnail,
+                    )
+                case Type<MetadataViews.HTTPThumbnail>():
+                    return MetadataViews.HTTPThumbnail(
+                        uri: self.thumbnail,
+                        mimetype: "image/jpeg",
```
turbolent, line 56:
```suggestion
                        mimetype: "image/jpeg"
```

contracts/MetadataViews.cdc
```diff
@@ -29,4 +29,55 @@ pub contract MetadataViews {
...
+
+    // HTTPThumbnail returns a thumbnail image for an object.
+    //
+    pub struct HTTPThumbnail {
+        pub let uri: String
+        pub let mimetype: String
+
+        init(
+            uri: String,
+            mimetype: String,
```
turbolent, line 57:
```suggestion
            mimetype: String
```

contracts/MetadataViews.cdc
```diff
@@ -29,4 +29,55 @@ pub contract MetadataViews {
...
+    // rather than a direct URI. A client application can use this CID
+    // to find and load the image via an IPFS gateway.
+    //
+    pub struct IPFSThumbnail {
+        pub let cid: String
+        pub let mimetype: String
+
+        init(
+            cid: String,
+            mimetype: String,
```
turbolent, line 77:
```suggestion
            mimetype: String
```

bjartek:
@niveda-krish Personally I think that representation is a little bit to thick. It will also have lots of nils since many people to not have token_uri and others have token_uri but no name and description on chain. Splitting those out into two different views is preferreable if you ask me. 

Also if you have more Media then the thumbnail they should be exposed a a [Media] type or something like that.

bluesign:
I agree with @bjartek on technical sides of the things, display with common properties.

But I think there is also a business side of the things, as I observed. I have strong feelings against but it is the reality. 

I think dapps would like to approve places where artwork can be presented. Should we consider this moving forward? 

You want your item listable on blocto bay, you implement bloctoDisplay metadata. Or if you want it to be visible on Viv3, you implement viveDisplay etc 

If restricting IP to be seen on only allowed websites, I think this can be useful standard to implement. ( in this bloctoDisplay, viv3Display or versusDisplay there should be royalty etc also )

If we go with generic display, can we say that I am providing this gives people right to show on any website? I think as licensing is complicated subject, I am not sure, though there is no ToS on blockchain. 

PS: technically I am in favor of single Display struct, but if it will slow down adaptation for licensing concerns, I think we need to implement something with more coverage. 

Will TopShot and NFL support metadata for example?

bjartek:
If you have ToS I think you would be better off not implementing Display or maybe just a Display with what you can view for free and then having another one that solutions that have signed the ToS can use. Or even have an authenticated ERC721 api behind that to solve those scenarios. 

Using seperate Display struct is not something i would recommend at all. It would require frequent contract upgrades and that is not trivial right now. We need ease of use and standardization now, not splitting things up even more.

rambhask:
Just echoing a minor thought about having an IPFS specific thumbnail - this feels like something that can be inferred from the url as @bjartek suggested. Also, I know some projects prefer to pass in a link to their own ipfs gateway (like pinata) and others have preferred Arweave of late due to their pay once, store forever model.

niveda-krish:
Alternate proposed views based on @bjartek's comments:

**View 1: (off-chain metadata)**
token_uri (points to IPFS, Arweave, any other url with metadata).

**View 2: (on-chain metadata)**
name
description
source URL + mimetype
thumbnail URL + mimetype
source
attributes (name-value pairs)

btspoony:
> Just echoing a minor thought about having an IPFS specific thumbnail - this feels like something that can be inferred from the url as @bjartek suggested. Also, I know some projects prefer to pass in a link to their own ipfs gateway (like pinata) and others have preferred Arweave of late due to their pay once, store forever model.

Agree, It’s weird to use `IPFSThumbnail` struct for projects using Arweave

psiemens:
> View 1: (off-chain metadata)
token_uri (points to IPFS, Arweave, any other url with metadata).

@niveda-krish Yeah I'm all for this. I've been thinking about proposing a separate PR with an `EIP1155JSON` view that looks just like that, based on the [EIP 1155 standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1155.md#erc-1155-metadata-uri-json-schema).

psiemens:
> Agree, It’s weird to use IPFSThumbnail struct for projects using Arweave

Well in this case I think there'd be a separate `ArweaveThumbnail` view.

I'm a bit worried about referencing IPFS and Arweave content with just a URL. That's my main hesitation with the approach above. I like the idea of using a dedicated type to tell a client app that the content lives in IPFS, Arweave or on a plain web server.

For example, I don't want to accidentally encourage developers to force apps to use a specific IPFS or Arweave gateway. If they _want_ to do this, they can use `HTTPThumbnail`. IMO it's safer to only return the content hash / CID and let the client choose their preferred gateway. A big benefit of IPFS/Arweave is that files are redundantly available through multiple public gateways.

I know that IPFS has the `ipfs://` prefix, but I guess I just don't like the idea of us resorting to prefix matching to distinguish types, _especially_ if it forces developers to do string processing in a Cadence script.

psiemens:
> I think dapps would like to approve places where artwork can be presented. Should we consider this moving forward?

@bluesign I probably haven't thought about this as much as others, but I don't think we should try to enforce IP restrictions on chain. 

If an application _really_ wanted to display an NFT that was only approved for Blocto Bay, they could just read the `bloctoDisplay` struct. Everything is public, so I'm not sure it's worth the effort to split things up.

bjartek:
> Well in this case I think there'd be a separate ArweaveThumbnail view.

I would much rather have a field in the Display struct that said what protocol this was rather then it beeing a seperate View. Note that this is only in the disaply struct thought. In the Media view I am all for specific implementations for different storage technologies. Like for Versus we need a Media View that does allows lazy loading of the content since we store it onChain and resolving it everytime you fetch that VIew is not desierable.

joshuahannan:
Niveda’s proposal seems to make sense to me, though I think that `source` could just use `getType` and I am undecided about whether having a `{String:String}` dictionary in there is the right call. 

@psiemens What are the arguments in support of keeping the views separate?

dete:
Let's remember that we need to balance the needs of today with the needs of tomorrow.

My assumption is that, in future, we will see more and more NFT projects choosing to make their images and other display data (videos, 3D models) available via decentralized systems. In fact, I would even suggest that we, as a community, should do everything we can to encourage that wherever practical. (And yes, this is a discussion we've had with both the NFL and NBA, and decentralized storage is a direction they've agreed on, although there's no specific timetable set yet.)

A URL string is easy. A URL string is comfortable. But it also encourages HTTP as the default. But HTTP is fragile. I want Flow NFTs to be something you can confidently pass down to your grandchildren in your will, and HTTP links don't qualify. Not even HTTP links to an IPFS gateway. (Unless you write a parser to extract the IPFS hash from the link, which is ugly!)

Of course we need to support HTTP, there are too many projects, Dapper's included, that have used this well-established technology for bootstrapping. But if we just leave the `image` field as a string, I fear that HTTP will quickly become the only safely-supported option.

By separating the thumbnail into its own view, with a different view for each transport type, we make it much explicit when new transport types are added. We also allow projects to support multiple transport types, which means there is no downside to support IPFS today, even if more marketplaces will only use HTTP in the short term, and ARWeave becomes the new default after that. You can programatically create an HTTPThumbnail view from your IPFSThumbnail or ARWeaveThumbnail view in your NFT class using your favourite gateway, if you wanted.

Basically, this comes right back to the reason why we created Metadata as a series of Cadence Types instead of a simple string-string dictionary in the first place: It makes the data much more self-describing, it makes it much easier for _smart contracts_ to use metadata in a meaningful way (Maybe I only want to accept NFTs that have an IPFS image view in my collection…), and converting Cadence types into a string-string dictionary for those places where it makes more sense off-chain is trivial.

In short, let's make sure we're designing for the future, and – in particular – a future where smart contracts are even more complex and interesting than they are today!

bjartek:
From my limited experience with IPFS, does it scale for quickly showing dozens of thumbnailed items? Blocto bay is using a CDN to host their images it looks like. I really want to be proved wrong here, but for a low-res thumbnail preview of an NFT i think CDNs over http will be the choice in the near future. 

I can really see your point though @dete. What about the following
```
pub struct interface Media {
		pub fun data() : String
		pub let mediaType: String
		pub let protocol: String

	}

	/// IPFS specify media that holds the CID as a field
	pub struct IPFSMedia : Media {
	        pub let mediaType: String
		pub let protocol: String
		pub let cid:String

		init(cid:String, mediaType: String) {
			self.cid=cid
			self.protocol="ipfs"
			self.mediaType=mediaType
		}

		pub fun data() : String {
			return self.cid

		}
	}

	/// This is media that is kept as a String onChain.
	pub struct StringMedia : Media{
		pub let content: String
		pub let mediaType: String
		pub let protocol: String

		init(content:String, mediaType: String) {
			self.content=content
			self.protocol="onChain"
			self.mediaType=mediaType
		}

		pub fun data(): String {
			return self.content
		}
	}


	/// Generic Media representation that allow you to represent any kind of media you like
	pub struct HttpMedia : Media{
		pub let content: String
		pub let mediaType: String
		pub let protocol: String

		init(content:String, mediaType: String) {
			self.content=content
			self.protocol="http"
			self.mediaType=mediaType
		}

		pub fun data(): String {
			return self.content
		}
	}

	/// Generic Media representation that allow you to represent any kind of media you like
	pub struct GenericMedia : Media{
		pub let content: String
		pub let mediaType: String
		pub let protocol: String

		init(content:String, mediaType: String, protocol:String) {
			self.content=content
			self.protocol=protocol
			self.mediaType=mediaType
		}

		pub fun data(): String {
			return self.content
		}
	}
	
	pub struct Medias {
		pub let media : {String:  &{Media}}

		init(_ items: {String: &{Media}}) {
			self.media=items
		}
	}
```

Very similar to what Pete proposed. I still think that Display should have a thumbnail field but for me having it be an `&{Media}` is a compromise I am willing to make.

psiemens:
@bjartek I want to clarify a few things to make sure I understand your perspective correctly.

1. Your main concern is the complexity of resolving multiple views rather than one.

2. However, you're only concerned about this in the context of the `Display` struct. Based on this comment, it sounds like you still think there should be a way for NFTs to define a variety of media assets:

>  In the Media view I am all for specific implementations for different storage technologies.

3. So am I correct in understanding that in your opinion, the `Display` struct is a convenience view that summarizes data that may exist in other views? 

For example, in your version, an NFT could implement both `Display` and `IPFSMedia`, where `Display.thumbnail` may refer to the same image as `IPFSMedia` (or a downscaled version).

Please tell me if that interpretation is wrong!

bjartek:
1. I think it would be benefitical if one only had to resolve 1 view in order to create a listing a many resources, since the complexity can grow very fast. 

2. Totally. The Display view is one way of viewing the data, there should always be other more specific views for a given scenario. Like if you support an Artist/Artwork struct or a [Media] list or something like that. 

3. Yes.

You are totally correct.

psiemens:
@bjartek What is the benefit of the `Media` interface over separate view structs?

boczeratul:
IMO I think what's stored on-chain should only be the ground level information that serves as the source-of-truth of these items. So for me, having something like `ArweaveThumbnail` or `IPFSThumbnail` makes a lot of sense. 

The processed `HTTP` version of the images should exist on the layer of Alchemy API for easy access from dapps. BloctoBay does similar processing. We will cache the images defined in the metadata and process the thumbnails and store in our CDN for better scalability and faster loading. But I don't think it's good to condense this information into a string field in the on-chain metadata of NFT.

bjartek:
> @bjartek What is the benefit of the `Media` interface over separate view structs? 

I just added that to be able to have a Medias struct that can hold all your media with a name for it. Also supporting onchain art like flovatar and versus is best served with not returning all the data when resolving all your media since it will bog down everything.

psiemens:
Here's a summary of the qualities I think we collectively want from this standard:

1. A single summary of the basic data needed to render an NFT (e.g. in a list or wallet display)
2. A self-describing way to distinguish between media sources (a web server, IPFS, Arweave, etc)
3. Flexibility to support additional storage solutions in the future
4. Ability for NFTs to define more than one media source (i.e. an NFT can reference the same image in a web server and IPFS)

## Option 1

A single `Display` struct with a string thumbnail.

- ✅ 1, 3
- ❌ 2, 4 (a `String` is not self-describing)

```swift
pub struct Display {
    pub let name: String
    pub let description: String
    pub let thumbnail: String
}
```

## Option 2

Remove thumbnail from `Display`, use separate thumbnail views.

- ✅ 2, 3, 4
- ❌ 1 (need to resolve 2 views to get full summary)

```swift
pub struct Display {
    pub let name: String
    pub let description: String
}

pub struct HTTPThumbnail {
    pub let uri: String
}

pub struct IPFSThumbnail {
    pub let cid: String
    pub let path: String?
}

pub struct ArweaveThumbnail {
    pub let id: String
}
```

## Option 3

Define a `Media` interface and use it for the thumbnail field.

- ✅ 1, 2, 3, 4

[See @bjartek's suggestion above](https://github.com/onflow/flow-nft/pull/55#issuecomment-1009396538)

## Option 4

- ✅ 1, 2, 3, 4

Define a lower-level `File` interface and use it for the thumbnail field. A variation of @bjartek's suggestion, but slightly more generic.

Notes:
- `protocol` is redundant given that the type itself tells the caller if a file is IPFS, HTTP, etc
- `init` functions are omitted for brevity

```swift
pub struct Display {
    pub let name: String
    pub let description: String
    pub let thumbnail: AnyStruct{File}
}

pub struct interface File {
  pub fun uri(): String
}

pub struct IPFSFile: File {
  pub let cid: String
  pub let path: String?

  pub fun uri(): String {
    if let path = self.path {
      return "ipfs://".concat(self.cid).concat("/").concat(path)
    }
     
    return "ipfs://".concat(self.cid)
  }
}

pub struct HTTPFile: File {
  pub let fileURI: String

  pub fun uri(): String {
    return self.fileURI
  }
}

pub struct Base64File: File {
  pub let content: String
  pub let mimeType: String

  pub fun uri(): String {
    // Example: "data:image/jpeg;base64,R0lGODdh"
    return "data:".concat(self.mimeType).concat(";base64").concat(self.content)
  }
}
```

### 4.1 - Examples

These are out of scope for this PR, but `File` could easily be used inside other media views.

I don't want to over-solve this part now; just showing examples of how flexible `File` can be.

Maybe an image can have optional text (e.g. for accessibility / screen readers):

```swift
pub struct Image {
  pub let image: AnyStruct{File}
  pub let description: String?
}
```

Here's how you could group images together (drawing inspiration from Bjarte):

```swift
pub struct Images {
  pub let images: {String: Image}
}
```

psiemens:
@bjartek @niveda-krish @btspoony @boczeratul What do you think of option 4 above?

bjartek:
Nr 4 looks awesome too me. Like the composability of that in 4.1 aswell. 

What about identity of resource (id,uuid,type and discriminator and sourceUri? Is that maybe seperate views?

psiemens:
> What about identity of resource (id,uuid,type and discriminator) and sourceUri? Is that maybe seperate views?

Source URI is in a separate view, as discussed above. I'll open a PR for that as soon as we merge this one.

> id,uuid,type and discriminator

ID and UUID should aways come directly from the `NFT` resource. A view could lie about either of those fields.

I think we should try to encourage `nft.getType()` whenever possible, but I am also aware of use case where multiple "collections" share a single contract. However, I think that should be solved with a separate view.

bjartek:
Works for me. :)

niveda-krish:
Option 4 looks great to me too - I'm guessing metadata will be in a different view?

bjartek:
@niveda-krish are you talking about an arbitrary {String:String} metadata or something with more context? I see why you would want just a generic Map for that but I think having more context to the values would give a lot more value. 

As an example both Gooberz, Ballerz, Flovatar consist of multiple parts. Having some way to describe those parts in a standard way would be way nicer then everybody just having a String:String dictionary with the parts and values in them.

psiemens:
I think it's valid to have a `{String: String}` map when needed, but would suggest that as a separate view.

psiemens:
Thanks everybody! I've updated this branch with a final version of option 4 above.

bjartek (approved):
This looks very good to me!

psiemens:
I'm going to wait **24 hours** before merging and deploying this change to give everybody a chance to review the code and voice any concerns.

I've also implemented these views in the [Kitty Items demo contract](https://github.com/onflow/kitty-items/blob/metadata/cadence/contracts/KittyItems.cdc#L132-L150), which may be a helpful example for people evaluating this standard.

Cheers!

rheaplex (approved):
❤️ Looks good! 😎

ssingal05:
Sorry, I am a little late to the party here and am now learning about the efforts the community has been making toward defining NFT metadata standards for Flow. My understanding is that we have an interface called `MetadataView.Resolver` that we want all NFT projects to implement. I wanted to propose an alternative approach which I think can further support the themes of "decentralization" and "composability" we see across blockchain dapps.

My immediate concerns are

- How will this practically affect past NFT adopters who will now have to develop solutions to make their project more compatible with this new metadata ecosystem that they are defining?
- I do not think that new NFT projects should be penalized for forgetting to extend `MetadataView.Resolver` but, rather, should be allowed to join the metadata ecosystem whenever and however they want
- This seems to disallow an NFT project from showing two different versions of `Display` based on the consumer

I also think that we are missing out an a potential community building/sourced opportunities. To better illustrate what I mean, I want to propose an "Adapter"-based solution. Literally speaking, these should be independent (Flow) resources that have the ability to take an NFT and adapt it to fit an NFT metadata definition (Like `Display` or `IPFSFile`). When web developers are building solutions for their sites, they'll look to the contracts to see what fits the requirements for their website (they want a `Display` because they just care about getting a simple thumbnail to display on their site) and then they'll look to "providers" that provide an adapter for the specific type. Using this strategy

- Creators can focus on making their NFTs contain exactly what they want the NFT to represent. It'll be up to them to provide enough information to allow more adapters to use them properly
- Metadata adapters can be completely community led! This relieves the NFT creator from having to provide metadata support at all! They can rely on the community to do it if they want, or do it themselves if they prefer that, or a combination of both (because now we can have two different `Display`s representing an NFT)
- Previous NFT developers can still take advantage of this fully!

We should be able to build standards and conventions to make this easier to work with for developers, but I'd like to hear any feedback on this approach.

pgebheim:
I haven't been a part of this direction but why do we have specific http/ipfs parts of this instead of using URIs which would allow for this to be extensible to other other storage formats?

`ipfs://<cid>` is the standard uri format for a particular piece of ipfs content.

A similar structure can be made for arweave and we already have it for http. This allows view functions to describe how the content is accessed instead of having it be on chain separately per format.

rheaplex:
> * How will this practically affect past NFT adopters who will now have to develop solutions to make their project more compatible with this new metadata ecosystem that they are defining?

They can update their code to support it. This is the advantage of a function-based interface.

> * I do not think that new NFT projects should be penalized for forgetting to extend `MetadataView.Resolver` but, rather, should be allowed to join the metadata ecosystem whenever and however they want

Absolutely.

> * This seems to disallow an NFT project from showing two different versions of `Display` based on the consumer

By the consumer do you mean the caller? That sounds interesting. Do you have an example of this being used?

> To better illustrate what I mean, I want to propose an "Adapter"-based solution. 

We have a proposal for a "Converter" based solution that is similar, and that complements the View based system described here.

rheaplex:
> I haven't been a part of this discussion so far, so apologies if I missed the reasoning, but why do we have specific http/ipfs parts of this instead of using URIs which would allow for this to be extensible to other other storage formats?

I'm team URI, hence the base interface function that returns a URI string. 😺

But a string doesn't give us the power of the Cadence type system. The reason for using the Cadence type system in the Metadata system is to allow people to express information more richly than they can using strings - and to be clear we use strings wherever they are appropriate, there's nothing wrong with strings.

If a project front-end wants to use http rather than ipfs, it can search for that type. And it can parse the information about that url carefully. If it doesn't care, it can just get the first File type it can and then call the uri function on it.

Separating out the components also allows us to optimize various common cases and make them simpler to update if needed - a root URI for http, or a directory CID for ipfs.

So I believe we have a nicely balanced approach here, but if you have criticisms or concerns beyond this do let us know.

ssingal05:
Thanks for the reply @rheaplex .

> They can update their code to support it. This is the advantage of a function-based interface.

Ah can you elaborate on this? For example,  the Top Shot developers would be able have to their currently deployed contract to extend `MetadataViews.Resolver`? My understanding was that contracts should be immutable as a way of reassuring users that code changes would not affect them in the future.

> By the consumer do you mean the caller? That sounds interesting. Do you have an example of this being used?

I think I could have worded that better, but what I mean is imagine you had an NFT representing a movie (MovieNFT), and you have a metadata standard representing a "Rating" (name and number rating). Rotten Tomatoes and IMDB could spawn their own metadata converter instances that resolve a movie to their own version of what the rating should be. The client (web browser) would pick the account address + storage path of the metadata converter they want to support on their website. They could decide to support them both, or they could look to an off-chain or on-chain solution that can help aggregate these rating providers. I don't have a particular example, but I hope this is able to illustrate what I think is a valid use case.

> We have a proposal for a "Converter" based solution that is similar, and that complements the View based system described here.

I would like to take a look if you have a link. I am not sure if it is a good idea to fragment metadata fetching strategies like this - I feel like we should prioritize developer experience, and having more options might mean more time needed to get tooling ready for callers, especially when I think the adapter/converter pattern can accomplish the current View pattern can do plus more.

briandilley:
@psiemens I think that `HTTPFile` should have an optional `hash` and `hashAlgorithm` (both strings) on it (like the example in the NFT MetadataSpec `URLWithHash`)

briandilley:
+1 for @psiemens suggesting above (#4)

briandilley:
I'd also like to add that `HTTPFile` should probably look like this:

```
pub struct HTTPFile: File {
  pub let fileURI: String
  pub let hash: String?
  pub let hashAlgorithm: String?

  pub fun uri(): String {
    return self.fileURI
  }
}
```

This way files can (optionally) be verified on HTTP as well.

contracts/MetadataViews.cdc
```diff
@@ -29,4 +29,86 @@ pub contract MetadataViews {
         pub fun borrowViewResolver(id: UInt64): &{Resolver}
         pub fun getIDs(): [UInt64]
     }
+
+    // Display is a basic view that includes the name and description
+    // of an object. Most objects should implement this view.
+    //
+    pub struct Display {
+        pub let name: String
+        pub let description: String
```
joshuahannan, line 38: Should we include comments about each field do describe what they should/could contain?

joshuahannan (approved):
Looks good to me! I just think we need more documentation/comments for it that describe the individual fields and what they should represent

niveda-krish (approved):

psiemens:
> I'd also like to add that HTTPFile should probably look like this

Thanks @briandilley, that's a good use case.

Could we implement that as a separate `File` implementation?

I'd like to avoid having optional arguments in the `HTTPFile` initializer. It could be confusing for developers who _don't_ need a content hash and could be misused by those who do (e.g. what happens if `hash` is passed but `hashAlgorithm` is nil?).

```swift
pub struct VerifiableHTTPFile: File {
  pub let url: String
  pub let hash: String
  pub let hashAlgorithm: String

  init(url: String, hash: String, hashAlgorithm: String) {
    self.url = url
    self.hash = hash
    self.hashAlgorithm = hashAlgorithm
  }

  pub fun uri(): String {
    return self.url
  }
}
```

psiemens:
Thanks @joshuahannan. I updated the `Display` view with comments.

briandilley:
@psiemens 👍 that makes sense to me.

poiati (approved):

psiemens:
Thanks everybody! I'm going to merge this PR in its current state and then will deploy the new `MetadataViews` contract.

@ssingal05 Sorry to cut your thread short, but given that your points are more related to be base view resolver pattern, it'd be great to continue that discussion in a separate issue.

turbolent:
Cadence provides a [built-in `HashAlgorithm` type](https://docs.onflow.org/cadence/language/crypto/#hashing). Using it would make this more type-safe.

ssingal05:
> Thanks everybody! I'm going to merge this PR in its current state and then will deploy the new `MetadataViews` contract.
> 
> @ssingal05 Sorry to cut your thread short, but given that your points are more related to be base view resolver pattern, it'd be great to continue that discussion in a separate issue.

Hey @psiemens - Frankly, I later realized that this PR was probably not the appropriate place for this discussion - that's my bad! Rhea mentioned a complementary converter pattern so it sounds like I'm not voicing any new concerns anyways. Thanks for your work here!

psiemens:
> Cadence provides a built-in HashAlgorithm type. Using it instead of a plain string would make this more type-safe.

@turbolent Good flag. I'm going to open a separate PR for `VerifiableHTTPView` so will account for that there!

briandilley:
> > Cadence provides a built-in HashAlgorithm type. Using it instead of a plain string would make this more type-safe.
> 
> @turbolent Good flag. I'm going to open a separate PR for `VerifiableHTTPView` so will account for that there!

I originally thought the same thing, but what if the hashing algorithm used is not supported by cadence?

PR #67, opened 2022-03-10
Move royalties to metadata views and cut down on extended utils 

joshuahannan:
* Put the Royalty view in the main `MetadataViews` contract where it will end up
* Modified a lot of the comments to be more succinct, have better grammar, etc
* Cut down the extended royalty functions to just two main utility functions that can potentially go in the metadata views contract


joshuahannan:
Thanks @bluesign! Yeah, that is just a potential utility function that we might include. Still haven't tested it yet, so it will likely change, but we think we should make it as easily as possible for marketplaces to verify the type of royalty receivers

contracts/ExtendedRoyaltyUtils.cdc
```diff
@@ -0,0 +1,55 @@
...
+    /// for better integration of the royalty standard with the marketplace
+    /// smart contract or provide help to provide better UI/UX.
+
+    //////////////////////////////////////////////////////////////////
+    /// Optional helper functions to facilitate the royalty integration 
+    //////////////////////////////////////////////////////////////////
+
+    /// Allow the sale lister to validate whether the beneficary of the royalty
+    /// is using the compatible capability with what buyer accepts.
+    pub fun checkWalletType(royalties: Royalties, typeToCheck: Type): Bool {
```
bjartek, line 16: This will not work with wallet abstractions. Personally I think that when we have default impl in place there should be a new method in FT<Receiver> That checks what Types that wallets accept. It can then be defaulty implemented to return a sane value.

bjartek (approved):

contracts/MetadataViews.cdc
```diff
@@ -129,4 +131,71 @@ pub contract MetadataViews {
...
+    *
+    */
+
+    /// Interface to provide details of the royalty.
+    pub struct Royalties {
+
+        /// Array that tracks the individual royalties
+        access(self) let cutInfos: [Royalty]
+
+        pub init(cutInfos: [Royalty]) {
```
bjartek, line 151: personally I do not see much use for labels in this case when there is a single paramter and it is very obvious what it is.

bjartek (approved):

satyamakgec (approved):

bjartek:
I just changed to using this royalty representation in a project I am doing compared to my old one (that looked very similar) it works as intended.

contracts/MetadataViews.cdc
joshuahannan: Good point. I made it optional

PR #68, opened 2022-03-15
correct the url for the given paths

Nasir-Ali-Shah:
The given URLs in the file were not correct, it was redirecting to a 404 page, update that to the MetadataViews contract.


joshuahannan (approved):
Thank you for catching this!

PR #69, opened 2022-03-16
[WIP]: Addition of Royalty view in the ExampleNFT and MetadataViews

satyamakgec:


joshuahannan (commented):
Looks good so far! I just remembered that we also said that we want to have people use a separate fungible token receiver from their main one, so we should define a getter function in the MetadataViews contract that returns a standard `PublicPath` for where the capability should be. This can't be a field on the contract for upgrade reasons, so it'll need to be gotten from contract storage, similar to [what we do in the staking contract.](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1247-L1249)

And then we'll need to update the transactions to include ones that create this new link at the path for the royalty receiver. and also update the minting transaction to get the receiver from this path

lib/go/test/metadata_test.go
```diff
@@ -25,26 +27,45 @@ func TestGetNFTMetadata(t *testing.T) {
...
 	const (
-		name = "Example NFT 0"
+		name        = "Example NFT 0"
 		description = "This is an example NFT"
-		thumbnail = "example.jpeg"
+		thumbnail   = "example.jpeg"
 	)
-	
+
+	cut1, err := cadence.NewUFix64("0.25")
```
joshuahannan, line 53:
```suggestion
        // Add two new royalties to the minted NFT
	cut1, err := cadence.NewUFix64("0.25")
```

lib/go/test/metadata_test.go
```diff
@@ -78,4 +99,24 @@ func TestGetNFTMetadata(t *testing.T) {
...
+	var results map[string]interface{}
+
+	// Unmarshal or Decode the JSON to the interface.
+	json.Unmarshal([]byte(royalties), &results)
+
+	// for key, value := range results {
+	// 	// Each value is an interface{} type, that is type asserted as a string
+	// 	fmt.Println(key, value.(string))
+	// }
+}
```
joshuahannan, line 115: Remember to remove these before merging

lib/go/test/metadata_test.go
```diff
@@ -78,4 +99,24 @@ func TestGetNFTMetadata(t *testing.T) {
...
+
+func toJson(t *testing.T, target cadence.Value) string {
+	actualJSONBytes, err := jsoncdc.Encode(target)
+	require.NoError(t, err)
+	actualJSON := string(actualJSONBytes)
+	return actualJSON
```
joshuahannan, line 121: Can you put this in `test.go`?

joshuahannan (commented):
The changes look good! I think the next step would be to add a section in the README that briefly describes royalties, then we need to find a place in the docs site to add docs and tutorials for setting up and using royalties.

joshuahannan:
I added some docs in the README and cleaned up a couple other things. If nobody else has any reviews they want to give, I'm gonna merge soon!

joshuahannan (approved):

Issue #59, opened 2022-02-04
External Domain view Url


bjartek:
Both .find and alchemy have the concept of a url to your NFT on the web, I think having a single view for this would make it a lot easier to interop between solutions. 

I propse the following simple view for this to be added to MetadataViews

```
pub struct ExternalDomainViewUrl {
  
  pub let url:String

  init(url: String) {
    self.url=url
  }
}
```

Issue #65, opened 2022-02-12
Media View


bjartek:
I suggest some Media views that add contentType meaning to the File struct interface. 

- Image
- Video
- Audio

Issue #56, opened 2022-01-05
Identity View


bjartek:
## Instructions

Add a Identity view so that resources can Identify who they are

### Problem

How do you know that you are buying a flowverse sock from the correct minter? Discovering this progrematically is very hard and almost impossible. 

Also the very useful uuid field is hidden from the NFT.

### Acceptance Criteria

Implement a View in the new NFT standard that expose some common information about an NFT, most of this can be default implemented in the contract, atleast there could be a method to generate it for you. 

The information I can think of is:

- id: AnyStruct, the id of this resource, the reson this is anyStruct is because the view is not tied to NFT and other resources can have other ids. (like .find names are identified by the name)
- type: Type, the Type of this resource
- uuid: the UIn64 uuid field
- discriminator: This is a text field to discriminate NFTS that are of the same Type but different origin
  
Adding this view to the Standard will and getting it adopted will make it easer for everybody to uniqely identify resources. The reson id/type is there is really to be able to map from the old pattern to the new pattern.

bjartek:
Come to think of it, a lot of the information here is already accessible runtime from a &{MetadataViews.Resolver} reference like id, type and uuid. So maybe this could just boil down to a `Minter`or  `Source` view that discriminates who really minted this, if the contract supports multiple minters.

bjartek:
Come to think of this even more, I do not belive id is present in `&{MetadatViews.Resolver}` so i still think this is a good idea.

hichana:
@bjartek this is an interesting nut to crack here. How do other chains solve this problem if at all? Or if there is no technical solution a given other chain, how does the community have some consensus on the correct minter?

bjartek:
Well on chains that are not resource oriented a shared contract would give bad resource isolatiob, but on flow resources are owned by users not contracts. So i guess the problem is not there.

hichana:
I was thinking like on an evm chain if someone were to clone and deploy a contract, who is to say which one is the 'correct minter'? Maybe that's what the 'discriminator' field here would do?

joshuahannan:
@bjartek Can you elaborate a bit more on the discriminator? I'm not sure how that is used to differentiate?

Also, what is stopping someone from faking the discriminator and the nonce?

for `path`, you're saying that if you had a random generic NFT, you could get the path from it to deposit it without knowing the actual contract it came from?

bjartek:
Discriminator could be used to discriminate things like

 - the minter of a starly card
 - the kind of nft in gaia
[...]

PR #70, opened 2022-03-16
Added metadata view

bjartek:
I compiled a PR of many of the issues I have created in the nft-repo. I chose the ones that I think are least controversial. Almost all of these are optional.


bjartek:
I seem to have tagged #56 and not #65 in a commit

joshuahannan (commented):
Can you add some more comments to the views to explain them a bit more?

Also, I think that SerialNumber could probably be in a different PR because that one will probably warrant more discussion

bjartek:
Come to think of it, sometimes Video/Audio/Image is not enough. What if there is a new File type that suddenly requires a new wrapper. What if we just propsed Media that has a file and a contentType?

What particular views did you want to explain more? They are described in the corresponding issues?

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+        pub let file: AnyStruct{File}
+
+        init(_ file: AnyStruct{File}) {
+          self.file=file
+        }
+    }
+
+    // A license according to https://spdx.org/licenses/
+		//
+		//This view can be used if the content of an NFT is licensed.
```
joshuahannan, line 164: Formatting issues here too

joshuahannan (commented):
I think that some description of what file types you would expect in these would be useful
with specific examples

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+        pub let spdxIdentifier: String
+
+        init(_ identifier: String) {
+            self.spdxIdentifier = identifier
+        }
+    }
+
+    //A view to expose an url to this item on an external site
+
+		//This can be used in solutions like .find/blocto to know what should happen when you click on a link
```
psiemens, line 175:
```suggestion
    // A view to expose a URL to this item on an external site.
    //
    // This can be used by applications like .find and Blocto to direct users to the original link for an NFT.
```

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+    pub struct ExternalDomainViewUrl {
+        pub let url:String
```
psiemens, line 177:
```suggestion
        pub let url: String
```

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+
+        init(_ identifier: String) {
+            self.spdxIdentifier = identifier
+        }
+    }
+
+    //A view to expose an url to this item on an external site
+
+		//This can be used in solutions like .find/blocto to know what should happen when you click on a link
+    pub struct ExternalDomainViewUrl {
```
psiemens, line 176:
```suggestion
    pub struct ExternalURL {
```

Maybe just shorten this name?

contracts/MetadataViews.cdc
psiemens, line 164:
```suggestion
		// This view can be used if the content of an NFT is licensed. 
```

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
             return "ipfs://".concat(self.cid)
         }
     }
+
+
+    // A view to represent a File that is an image
```
psiemens, line 134:
```suggestion
    // A view to represent an audio file.
```

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+    pub struct Image {
+        pub let file: AnyStruct{File}
+
+        init(_ file: AnyStruct{File}) {
+          self.file=file
+        }
+    }
+
+
+    // A view to represent a File that is video
```
psiemens, line 144:
```suggestion
    // A view to represent a video file.
```

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,55 @@ pub contract MetadataViews {
...
+    pub struct Video {
+        pub let file: AnyStruct{File}
+
+        init(_ file: AnyStruct{File}) {
+          self.file=file
+        }
+    }
+
+    // A view to represent a File that is audio
```
psiemens, line 153:
```suggestion
    // A view to represent an audio file.
```

psiemens (commented):
Thanks for adding these!

bjartek:
Thanks for your suggestions @psiemens. I have changed to Media instead of the specific structs and fixed formating.

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,38 @@ pub contract MetadataViews {
             return "ipfs://".concat(self.cid)
         }
     }
+
+    // A view to represent Media, a file with an correspoiding mediaType.
+    // media-type comes on the form of type/subtype as described here https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
```
joshuahannan, line 134: Could we add this comment above the `mediaType` line instead of above the `Media` struct?

contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,38 @@ pub contract MetadataViews {
...
+    pub struct Media {
+        pub let file: AnyStruct{File}
+        pub let mediaType: String
+
+        init(file: AnyStruct{File}, mediaType: String) {
+          self.file=file
```
joshuahannan, line 140: Is there any sort of verification pre-condition we can do for `mediaType`?

joshuahannan (approved):

joshuahannan:
Closes #59

contracts/MetadataViews.cdc
bjartek: Yes

contracts/MetadataViews.cdc
bjartek: would this work?

```
 pre {             
  mediaType.utf8.contains("/".utf8[0]) : "Media type must be on the form type/subtype"
 }
```

Basically check the the media type contains a /.

joshuahannan: yeah, I think that works

joshuahannan:
I feel pretty good about these and we can probably merge soon and deploy them at the same time we deploy the royalties views. Do we think we need any tests for these or should we put any of them in the example NFT contract? 

@psiemens What do you think?

psiemens:
@joshuahannan @bjartek I'm good to deploy this as-is alongside the royalty view.

contracts/MetadataViews.cdc
joshuahannan: I agree it doesn't do much, but it at least will catch a few. But it does add three lines of code, which I guess might be more noise than it is worth

bjartek:
+1

joshuahannan:
@bjartek Can you pull the latest master, `make test`, then resubmit your branch?

bjartek:
Sure can do

bjartek:
Should be good now.

bjartek:
Done

Issue #58, opened 2022-01-19
MetadataViews.SerialNumber


poiati:
### Issue To Be Solved

Now that we have `MetadataViews.Display`, the next piece of data which is present in a significant portion of the NFT's out there is the so-called `Serial Number`. Unlike name and description, this information is not appliable to all tokens. That said, it's still something people care about. 

### Solution 1: MetadataViews.SerialNumber

Introduce the `MetadataViews.SerialNumber` view:

```
// ... MetadataViews Contract ...

pub struct SerialNumber {
  pub let serialNumber: UInt64
  pub let totalMinted: UInt64? // Not sure about this
}
```

PS: I'm still unsure if we should include the total number of minted NFT's. E.g: #1/200 (1 out of 200 copies). Mainly because this information is more likely to change.

### Solution 2: Serial number as part of the name / description

_This is not a solution, just an alternative._ 

If we don't want the Serial Number as first-class information within `MetadataViews` we can include it as part of the name/description in the `Display` view. It would be complex and error-prone if the client wanted to extract this information.

joshuahannan:
Do we think that the serial number can always be enforced to be a UInt64?

I like having a specific view for serial number, but I don't think it should be included in Display

bluesign:
I agree with @joshuahannan, I think `Display` should be as thin as possible.

poiati:
Yes, what I'm suggesting is a view apart from `Display`.

> Do we think that the serial number can always be enforced to be a UInt64?

I would say yes, @joshuahannan . What are your concerns? Is it about the size? Have you seen any other types being used? I see benefits from a dev perspective in having a homogenous representation. 

> @poiati did I understand your intention here or is this something else?

Yes, you nailed it @bjartek.

bjartek:
I think then that Solution 1 is the way to go, but I am not sure on the naming of the struct or the fields. I think `Editions` with the fields `edition` and `totalEdition`/`maxEdition` is more clear here. Just my 2 cents.

joshuahannan:
`UInt64` is totally fine with me. I just was thinking that there might be a chance that other projects would want something else, but I don't think it really matters. 

As for the name of the struct, I think `SerialNumber` is probably fine, but the second field should probably be something like `totalInEdition.` Though that is the first mention of Edition without any context, so maybe it isn't great.

SeanRobb:
Can we consider editions to have an `id`, `name`, and `description`?  This way I could see if NFTs are similar groupings of each other?

[...]

PR #75, opened 2022-04-11
Add Edition view

boczeratul:
Edition information is very useful in wallets and marketplaces. As more and more projects on Flow mint multiple items with the same name, description and media file, it is beneficial to put this into standard.

For projects that have items with editions, they should implement this `Edition` view.

Screen Shot 2022-04-11 at 9 02 06 PM
Screen Shot 2022-04-11 at 9 03 57 PM


contracts/MetadataViews.cdc
```diff
@@ -129,4 +129,26 @@ pub contract MetadataViews {
...
+
+        // The edition number of the object.
+        //
+        pub let editionNumber: UInt64
+
+        // The max edition number of this type of objects.
+        // 
+        // This field should only be provided for limited-editioned objects.
+        // 
+        pub let maxEdition: UInt64?
```
joshuahannan, line 147: Can you add a little bit more detail in the comments about what edition number and maxEdition mean? I assume edition number is like a "24 of 100" type thing and maxEdition is 100 in that case, but that should be crystal clear from the comments and docs here

joshuahannan (commented):
Just have the one comment so far! It looks good!

contracts/MetadataViews.cdc
boczeratul: updated.

psiemens (approved):
Thanks for adding tests and updating the examples!

satyamakgec (approved):
LGTM

albeethekid:
No comments, other than this look great!

joshuahannan:
@boczeratul We just merged another PR to master. Can you pull the latest master, `make test`, then push the resolved changes again?

poiati:
Please read the issue: https://github.com/onflow/flow-nft/issues/58. There is an extensive discussion about editions and we need more thoughts on it. It is currently blocking this PR. Thank you!

contracts/MetadataViews.cdc
```diff
@@ -238,4 +238,30 @@ pub contract MetadataViews {
...
+    // with the same name, description and media file. This information is 
+    // useful for wallets and marketplaces.
+    // 
+    pub struct Edition {
+
+        // The edition number of the object.
+        //
+        // For an "24 of 100 (#24/100)" item, the editionNumber is 24. 
+        //
+        pub let editionNumber: UInt64
```
austinkline, line 252: We already know that the view is about editions, what's the reasoning for repeating that again with each variable name?

albeethekid (approved):

bjartek:
The discussion in #58 has concluded so I think this can be merged?

joshuahannan:
no, I'm working on a more sophisticated version of it that I'll have ready relatively soon

bjartek:
I might have misunderstood something, but why do we need sophistication here? When we have Serial what do we need then except Edition(number:total)?

joshuahannan:
A project might have multiple editions that it is a part of, so the view should probably support that. Something like this:
```
pub struct Edition {

        /// An arbitrary-sized list for any number of editions
        /// that the NFT might be a part of
        pub let infoList: [EditionInfo]

        init(_ infoList: [EditionInfo]) {
            self.infoList = infoList
        }
    }

    /// Edition information for a single edition
    pub struct EditionInfo {

        /// The name of the edition
        /// For example, this could be Set, Play, Series,
        /// or any other way a project could classify its editions
        pub let name: String?

        /// The edition number of the object.
        ///
        /// For an "24 of 100 (#24/100)" item, the number is 24. 
        ///
        pub let number: UInt64

        /// The max edition number of this type of objects.
        /// 
        /// This field should only be provided for limited-editioned objects.
        /// For an "24 of 100 (#24/100)" item, max is 100.
        /// For an item with unlimited edition, max should be set to nil.
        /// 
        pub let max: UInt64?

        init(name: String?, number: UInt64, max: UInt64?) {
            self.name = name
            self.number = number
            self.max = max
        }
    }
```

joshuahannan:
@bluesign I think that information could be included in the name, right?

bjartek:
Is it neccesary to pack a list if editions inside another type? A view of [Edition] is totally valid?

joshuahannan:
Just following the same pattern that we used for royalties. When we say that the function can resolve a specific type, I like for the resolver to return an object of the actual type, not an array of objects of that type

bjartek:
If we want to follow the example of Royalties can we then name the view Editions and the thing inside Edition?

joshuahannan:
Good suggestions! New PR is in https://github.com/onflow/flow-nft/pull/88

bluesign:
> @bluesign I think that information could be included in the name, right?

I am not sure tbh, Name looks like 'Set' here, it is more like a value ("Young Stars")

PR #77, opened 2022-04-20
MetadataViews.ResolverCollection link updated.

Nasir-Ali-Shah:
MetadataViews.ResolverCollection link was missing in the init of the contract while linking Collection of the contract.


Nasir-Ali-Shah:
> Thank you! Can you also pull the latest changes from master and push again so the conflicts are resolved?

Thanks for the feedback, I have updated the readme file.

PR #78, opened 2022-04-20
update royalty setup transaction to correct parameter placement

joshuahannan:
Before, the storage path parameter was in the `prepare` parameter list, when it should have been in the `transaction` parameter list. Updated the transaction to move that

Also, removed the FungibleToken.Vault concrete type specification


satyamakgec (approved):
LGTM

Issue #62, opened 2022-02-04
contract/collection level metadata


btspoony:
## Issue To Be Solved

At present, when Flow wallets or DApps integrate an NFT Collection, what the basic information of NFT collections (like display name, description, etc.) needs to be collected case by case.

Have we considered proposing a `FLIP` that implements a standard  for the definition of contract or collection level metadata?  
Reference: https://docs.opensea.io/docs/contract-level-metadata

If such an standard exists, the work will be handed over to the NFT contract developers who will implement Views like `GeneralDisplay`, `BloctoDisplay`, `AlchemyDisplay`, `NewWalletDisplay` with their collection level information(contract name, display name, description, banner image, etc.).

bjartek:
This is what alchemy needs for their impl so that should be considered. https://github.com/alchemyplatform/alchemy-flow-contracts/blob/main/src/cadence/mainnet/getNFTs.cdc#L90

lmcmz:
Yeah, this a good idea. I also made a similar suggestion in their repo, but haven't got any reply on this yet.  :/
https://github.com/alchemyplatform/alchemy-flow-contracts/issues/35

joshuahannan:
Can you elaborate on what kind of info would be included in this view? Seems to me like most of this is already included in other views

btspoony:
> Can you elaborate on what kind of info would be included in this view? Seems to me like most of this is already included in other views

This does not refer to NFT information, but to **NFT collection** information. 
For example:
- **name** of the collection
- **description** of the collection
- **logo image** of the collection
- **official website** of the collection
- **storage_path** of the collection
- **public_path** of the collection
- **public_collection_name** of the collection

which can be used in https://github.com/alchemyplatform/alchemy-flow-contracts/blob/e44cea7ebe90d1b289f999c2b40c4ed4393ffd0a/src/cadence/mainnet/getNFTs.cdc#L117
and also can be used for DApps displaying for the NFT collection (not each NFT)

joshuahannan:
If you have a reference to a collection, isn't a lot of this information already available? You had to use the name and paths to get it already. 

Also, how would you propose we integrate this with the standard? This doesn't fit with the metadata standard because that is for individual NFTs. Would you propose adding another contract interface to implement?

btspoony:
> * 
> * **storage_path** of the collection
> * **public_path** of the collection
> * **public_collection_name** of the collection

[...]

PR #82, opened 2022-05-03
Add NFTCollection View

aishairzay:
Provides a view that enables storage and retrieval of NFTs without explicit knowledge being needed beforehand off-chain.

The view is at the NFT level and not the collection level because this allows for dynamic paths per NFT. This is essential to certain collections that may want to have multiple paths supported within a single NFT contract.

5 fields are expected to be implemented for each NFT as part of the NFTCollectionView:
* storagePath: Storage path where the relevant collection for storing this NFT should be located in an account storage
* publicPath: A public path where a public capability should be linked which provides access to the `publicCollectionLinkedType` below
* publicCollection: A type (normally restricted) that should be used when accessing this NFT. 
* publicCollectionLinkedType: A type (normally a restricted type) that should be linked on the public path.
* createEmptyCollectionFunction: A function that returns a `@NonFungibleToken.Collection` which is expected to be used at the aforementioned storage path to hold this NFT.

The above types will allow DAPPs to properly setup and read an NFT's collections without needing to explicitly *know* the NFT contract type.

One limitation is that using just the type is not sufficient to setting up a collection within a transaction because of the following cadence limitation: https://github.com/onflow/cadence/issues/1617
Once the above is resolved, it should be possible to use this view to fully setup a collection's storage and public capabilities without requiring to import the NFT. Until that is resolved, you may use the `publicCollectionLinkedType` type's `identifier` to retrieve all required imports to create a list of imports to create templated transactions using this view. The test in the exampleNFT showcases this.

Related issue: https://github.com/onflow/flow-nft/issues/62


contracts/ExampleNFT.cdc
```diff
@@ -60,6 +61,16 @@ pub contract ExampleNFT: NonFungibleToken {
                     return MetadataViews.Royalties(
                         self.royalties
                     )
+                case Type<MetadataViews.NFTCollectionView>():
+                    return MetadataViews.NFTCollectionView(
+                        storagePath: ExampleNFT.CollectionStoragePath,
+                        publicPath: ExampleNFT.CollectionPublicPath,
+                        publicCollection: Type<&ExampleNFT.Collection{ExampleNFT.ExampleNFTCollectionPublic}>(),
+                        publicCollectionLinkedType: Type<&ExampleNFT.Collection{ExampleNFT.ExampleNFTCollectionPublic,NonFungibleToken.CollectionPublic}>(),
+                        createEmptyCollectionFunction: (fun (): @NonFungibleToken.Collection {
```
bluesign, line 70: I think can assign ‘ ExampleNFT.createEmptyCollection’ here

aishairzay: That actually doesn't work, I don't know why though - this is what I get back with `ExampleNFT.createEmptyCollection`:
```
error: runtime error: invalid memory address or nil pointer dereference
```

bjartek: Would it be possible to have a path that is private and links to something that is a provider in adition to the standard restrictions? 

This would enable me to create a capability for say storefront or similar services from this information.

bjartek:
Would it be possible to have a path that is private and links to something that is a provider in addition to the standard restrictions? 

This would enable me to create a capability for say storefront or similar services from this information.

contracts/ExampleNFT.cdc
bjartek: Why are the paths in this example not including MetadataViews.ResolverCollection?

bluesign: @aishairzay this needs a bug report on cadence I guess, normally should work

aishairzay: I will add metadataviews.resolver and nft.receiver @bjartek

I think I left them out while testing to shorten the strings in the tests 😛

aishairzay:
> Would it be possible to have a path that is private and links to something that is a provider in addition to the standard restrictions?
I think a private provider path makes sense, do you have an opinion on if it should be optional or required? I'm leaning towards required, but I could imagine an argument for optional.

I also think it would need a corresponding type then which the private link is supposed to be borrowable or linked as.

bjartek:
> I also think it would need a corresponding type then which the private link is supposed to be borrowable or linked as.
> 
> 

Yes.

Are the paths validated at all? For this to make sense some restrictions should always be there IMHO.

aishairzay:
> Are the paths validated at all? For this to make sense some restrictions should always be there IMHO.

I'm interested to hear if you think some validation should be done at the `flow-nft` level @bjartek ?

I think the individual implementations that make use of these paths in transactions or contracts would be expected to decide themselves how to handle what may be an invalid path. This could mean re-linking capabilities or failing a transaction.

bjartek:
I think it would make sense to validate that it exposed the core NFT restrictions. So NFT.CP, NFT.Receiver and NFT.provider for private. 

And since this is a view and it is fetched using the standard both parhs should also expose MetadataViews.ResolverCollection.

austinkline:
> Provides a view that enables storage and retrieval of NFTs without explicit knowledge being needed beforehand off-chain.
>
> The view is at the NFT level and not the collection level because this allows for dynamic paths per NFT ...

Please correct me if I've missed something, but shouldn't we have some kind of default on the collection itself? Otherwise manual investigation is needed to understand individual assets and their paths which most of the time will line up. To me, we need some form of both since you don't always have the NFT. And you would need the NFT to ask for its path, right?

bjartek:
Will that no be kind of a chicken and egg problem @austinkline? You need a collection to get that view, and if you do not have one you need the view to create it?

Will there be a registry of nft identifier to NFTCollection/list of them somewhere?

austinkline:
> Will that no be kind of a chicken and egg problem @austinkline? You need a collection to get that view, and if you do not have one you need the view to create it?
> 
> Will there be a registry of nft identifier to NFTCollection/list of them somewhere?

Good question! So I'm coming from the use case of "tell me where this nft exists" when I don't have a reference to it yet. Think about how someone would build an indexer for flow right now. I might see a deposit event, but how do I borrow a reference to the collection to get the metadata for the nft that was deposited?

If I understand what you're asking, it's more about "how do I setup this collection" right?

aishairzay:
I agree with your points @austinkline - this PR will only solve just the angle of `I have an NFT how do i use it`, and separate efforts should address the other collection-level use-cases.

New PR is up - I added the requirements of types as you mentioned @bjartek

One thing i'm not sure of is the naming of the fields. If anyone has suggestions for change, please drop them here!

austinkline:
> I agree with your points @austinkline - this PR will only solve just the angle of `I have an NFT how do i use it`, and separate efforts should address the other collection-level use-cases.
> 
> New PR is up - I added the requirements of types as you mentioned @bjartek
> 
> One thing i'm not sure of is the naming of the fields. If anyone has suggestions for change, please drop them here!

Absolutely! Appreciate the context 🤝

bjartek:
I think the private path should require the paths the public one has aswell. If not is it impossible to get data from a private cap and imho it does not make sense.

aishairzay:
> I think the private path should require the paths the public one has aswell. If not is it impossible to get data from a private cap and imho it does not make sense.

Good point, added those in. I did not require that NFT.Receiver is a part of the private path, because that one didn't seem needed.

contracts/MetadataViews.cdc
```diff
@@ -238,4 +239,61 @@ pub contract MetadataViews {
...
+
+        // Public collection type that is expected to be linked at the aforementioned public path
+        // and provides sufficient access to standard functions (i.e. deposit + getIDs + borrowNFT)
+        // This is normally a restricted type with a single interface.
+        pub let publicCollection: Type
+
+        // Type that should be linked at the aforementioned public path. This is normally a
+        // restricted type with many interfaces. Notably the `NFT.CollectionPublic` and
+        // `NFT.Receiver` are required.
+        pub let publicLinkedType: Type
```
joshuahannan, line 267: I'm still a little bit confused on the difference between `publicCollection` and `publicLinkedType` and why we would need both of them. Can you provide more clarification?

lib/go/test/metadata_test.go
```diff
@@ -101,5 +101,80 @@ func TestGetNFTMetadata(t *testing.T) {
...
+	// Run a script to get the Display view for the specified NFT ID
+	//metadataScript := templates.GenerateGetNFTMetadataScript(nftAddress, exampleNFTAddress, metadataAddress)
+	/*metadataResult := executeScriptAndCheck(
+		t, b,
+		metadataScript,
+		[][]byte{
+			jsoncdc.MustEncode(cadence.NewAddress(exampleNFTAddress)),
+			jsoncdc.MustEncode(cadence.NewUInt64(0)),
+		},
+	)*/
```
joshuahannan, line 144: Can this be removed?

transactions/setup_account_from_nft_reference.cdc
```diff
@@ -0,0 +1,28 @@
+import NonFungibleToken from "../contracts/NonFungibleToken.cdc"
+import MetadataViews from "../contracts/MetadataViews.cdc"
+import ExampleNFT from "../contracts/ExampleNFT.cdc"
+
```
joshuahannan, line 4: Can you add comment at the top of this with a high-level description?

transactions/setup_account_from_nft_reference.cdc
```diff
@@ -0,0 +1,28 @@
...
+transaction(address: Address, id: UInt64) {
```
joshuahannan, line 5:
```suggestion
transaction(address: Address, publicPath: PublicPath, id: UInt64) {
```
Would it make sense to make the collection public path an argument to this also, so that when the other cadence feature is complete, we can do everything without having to import the `ExampleNFT` contract?

joshuahannan (commented):
Looks good! I just have a few comments

contracts/MetadataViews.cdc
aishairzay: The reason for both is for backwards compatibility with existing collections.

Many collections have not been linking NFT.CP because it was not enforced in the past, and only their specific NFT's CollectionPublic exposed by accounts that hold their collections.

Having this separate publicCollection which is guaranteed to work for the `read` use-case will allow DAPPs (like marketplaces) to read from collections in a standard way. And the `publicLinkedType` will serve the purpose of supporting setup or re-linking the public path in the proper way with standard collections.

joshuahannan: ok, that makes sense. Can you add a bit of that clarification in the comments as well?

contracts/MetadataViews.cdc
```diff
@@ -238,4 +239,63 @@ pub contract MetadataViews {
...
+        // `NFT.Receiver`, and `MetadataViews.ResolverCollection` interfaces are required.
+        pub let publicLinkedType: Type
+
+        // Type that should be linked at the aforementioned private path. This is normally
+        // a restricted type with at a minimum the `NFT.Provider` interface
+        pub let providerLinkedType: Type
+
+        // Function that allows creation of an empty NFT collection that is intended to store
+        // this NFT.
+        pub let createEmptyCollectionFunction: ((): @NonFungibleToken.Collection)
```
psiemens, line 277: Small nitpick, but I'd probably call this `createEmptyCollection` just so that it reads more naturally when invoked:

```cadence
let collection = data.createEmptyCollection()
```

The init parameter could still be `createEmptyCollectionFunction` though

psiemens (approved):

contracts/MetadataViews.cdc
aishairzay: I like that, updated to match the suggestion.

albeethekid (approved):

PR #83, opened 2022-05-09
Add NFTCollectionDisplay view to provide overview of an NFT's parenting collection

aishairzay:
Provides a view that allows an NFT to provide basic information about how to display its parenting NFT Collection.

 Similar to the collection view here: https://github.com/onflow/flow-nft/pull/82, this view provides a way of going from already having an NFT reference to retrieving the collection information that is parent to that NFT. For that reason, this has specifically been named `NFTCollectionDisplay` to not be confused with the `Display` view which is a generic Display view for any type of resource.

Related issue: https://github.com/onflow/flow-nft/issues/62


transactions/scripts/get_nft_metadata.cdc
```diff
@@ -22,28 +28,44 @@ pub struct NFT {
...
+        self.collectionDescription = collectionDescription
+        self.collectionExternalURL = collectionExternalURL
+        self.collectionSquareImage = collectionSquareImage
+        self.collectionBannerImage = collectionBannerImage
     }
 }
 
+pub struct NFTCollection {
+    
+}
```
joshuahannan, line 67: What is this?

aishairzay: My mistake - was thinking of reformatting things, and left that over. Removing it.

satyamakgec (approved):

bjartek (approved):

bjartek:
This is really needed, and since it is no the NFT level it works perfectly for multi-minter contracts aswell.

contracts/MetadataViews.cdc
```diff
@@ -298,4 +297,33 @@ pub contract MetadataViews {
             self.createEmptyCollection=createEmptyCollectionFunction
         }
     }
+
+    // A view to expose the information needed to showcase this NFT's collection
+    //
+    // This can be used by applications to give an overview and graphics of the NFT collection
+    // this NFT belongs to.
+    pub struct NFTCollectionDisplay {
```
austinkline, line 305: Should we add a "social links" field here as well? So then a collection can advertise its discord, twitter, etc from a central place? If so, I'd just make it a `{String: String}` and let folks specify

austinkline (approved):

Issue #74, opened 2022-04-03
NFTView


austinkline:
### Issue To Be Solved
A metadata view that encompasses all data of an NFT resource is needed, otherwise we have to rely on hand-made scripts like what's used in Alchemy's NFT endpoint which differs from what folks at .find return. These scripts shouldn't be needed, I should instead be able to ask for a standard view which would contain all information about an NFT that other platforms can then support. 

Separately, a complete view is necessary for dapps to support features on their platforms which require on-chain metadata. For instance, suppose there is a marketplace that wants to allow people to make global offers on nfts, allowing any nft to fill the offer so long as it has certain metadata key/value pairs.

### (Optional): Suggest A Solution
After some discussion with @bjartek, @bluesign, and amit (don't know their github handle), there are a few options for this that I see. First is that we could use the `DictView` as I have brought up in #73. This would be flexible enough to work, but some concerns were raised about its lack of explanation on the data being shown. 

Another option is to adopt the [OpenSea API Metadata Standard](https://docs.opensea.io/docs/metadata-standards) as the NftView. Here is what that structure looks like:
```
{
  "description": "Friendly OpenSea Creature that enjoys long swims in the ocean.", 
  "external_url": "https://openseacreatures.io/3", 
  "image": "https://storage.googleapis.com/opensea-prod.appspot.com/puffs/3.png", 
  "name": "Dave Starbelly",
  "attributes": [ ... ], 
}
```

And attributes are made up of:
```
{
  "display_type": "number",  # this one is optional
  "trait_type": "Generation", # this is the key name
  "value": 2
}
```

Using that as the model, our view would look something like this:
```
pub struct NftView {
    pub let description: String
    pub let externalUrl: String
    pub let image: String
    pub let name: String
    pub let attributes: [NftViewAttribute]
}

pub struct NftViewAttribute {
    pub let displayType: String?
    pub let traitType: String
    pub let value: AnyStruct
}
```

I think that we can do away with the things already covered in other views here. I'm just leaving them for reference and consistency with the returned format from OpenSea. You could strip away `description`, `externalUrl`, `image`, and probably `name` as well. In that case, we are just returning attributes in the format OpenSea returns them. This is a model that folks in the NFT community would be familiar with since that's how much of Ethereum's tokenURI results are formatted. Alignment with the broader NFT ecosystem would be a nice plus there.

[...]

Issue #60, opened 2022-02-04
Rarity View


bjartek:
Many different solutions on flow have the concept of Rarity. I think it would make sense to express this logic as a shared view to interop better between solutions. 

My naive implementation looks like this. 

```	// Would this work for rarity? Hoodlums, flovatar, Basicbeasts? comments?
	pub struct Rarity{
		pub let rarity: UFix64
		pub let rarityName: String
		pub let parts: {String: RarityPart}

		init(rarity: UFix64, rarityName: String, parts:{String:RarityPart}) {
			//TODO: pre that rarity cannot be above 100.0
			self.rarity=rarity
			self.rarityName=rarityName
			self.parts=parts
		}
	}

	pub struct RarityPart{

		pub let rarity: UFix64
		pub let rarityName: String
		pub let name: String

		init(rarity: UFix64, rarityName: String, name:String) {

			self.rarity=rarity
			self.rarityName=rarityName
			self.name=name
		}

	}```

joshuahannan:
@bjartek Can you elaborate more on what `rarity` and `RarityPart` means?

joshuahannan:
I was thinking it would be more like serial number and number in edition or something

bjartek:
My meaning for this view was to model rarity for 
 - starly cards
 - flovatars
 - goats
 - gooberz
 - ballerz

And all other projects that have nfts that have a rarity score. Some of the have parts that make up the score that can be included as well.

joshuahannan:
I'm still not sure what those fields actually represent though. Are they a standardized number scale or still specific to each project?

bluesign:
Here `RarityPart` made me thinking, do we have/need some generic `Parts` view ?  Not sure if having `X_Part` for X property of a `Part` is very good idea.

satyamakgec:
@bjartek the rarity view is to depict the property that is rare in the given NFT ? or it would be a score only?

bjartek:
In my eyes Rarity signals how rare the entire NFT is. For instance a Flovatar can be considered legendary if it used a legendary booster in its making. If you have multiple things in a rarity that contribute to its rarity score I would use rarityPart to signal that.

satyamakgec:
Ex - If an NFT has multiple parts let's say an NFT has some traits like cigarettes, Cap, Scarf etc. then those traits may have their own rarity, so those traits can be considered as part and provide the name and the rarity score for that trait?

joshuahannan:
[...]

PR #87, opened 2022-05-17
Add Attributes view

austinkline:
Add a new Metadata view for nft attributes. 

### Summary

Adds a new metadata view that can be used to obtain nft attributes. This approach follows the general guidelines of the OpenSea Metadata spec [here](https://docs.opensea.io/docs/metadata-standards#attributes) which should be a format much of the NFT ecosystem understands. 

Discussion about this view can be found in #74 as Attributes are part of an overall `NftView` which will come later.

### Tests
- One test has been added to evaluate the result of this new Attributes view.
- `make ci` passes


bjartek:
Would it be possible to add some helper functions to ease creating attributes. Like addStringValue(name:value) or something like that?

austinkline:
> Would it be possible to add some helper functions to ease creating attributes. Like addStringValue(name:value) or something like that?

Not opposed! Thinking about other usual types, do we want wrappers for all of them or just the more common of them? What comes to mind:
- String
- Bool
- UInt64
- Int64
- UFix64
- Fix64
- Address

bjartek:
I guess this depends on how we want the Attributes view to be used. Is it going to aggregate data from other views into this or are we supposed to add data to it directly. 

Come to think of it one issue that might occur with this view is that projects are going to use the same Attribute name with given types. So one project might have "rarity" as a number while another might have "rarity" as the string "Legenday". 

Not sure if that is a big problem or not though as you can know the type of the data.

austinkline:
> I guess this depends on how we want the Attributes view to be used. Is it going to aggregate data from other views into this or are we supposed to add data to it directly.
>
> Come to think of it one issue that might occur with this view is that projects are going to use the same Attribute name with given types. So one project might have "rarity" as a number while another might have "rarity" as the string "Legendary".
>
> Not sure if that is a big problem or not though as you can know the type of the data.

We can't predict or enforce all the kinds of attributes that will be used, so I don't think we can reasonably expect that a view per attribute can be made that will get rolled into this. Attributes are akin to a metadata dictionary like on TopShot plays:
https://github.com/dapperlabs/nba-smart-contracts/blob/master/contracts/TopShot.cdc#L147

The NftView which will be a separate PR is meant to roll-up the important views we want to show into one structure. Attributes are different though. In the end, attributes are what map to properties on sites like OpenSea:
image

I'd say that unbounded attributes shouldn't matter, I don't think it's up to us to define what each kind of attribute is allowed to be. Rarity makes sense as an example where things can be disjointed, but that very issue is why I think we should be hands off, who are we to decide what a valid value for rarity is? It could be a score of 1/100, 1/10000, 1/totalSupply, or just a string like TopShot's Tier. Evaluating the type of the attribute _should_ be enough to ensure that dapps can use it. 

Related to #60, if something is worthy of being elevated to a first-class citizen such as ExternalUrl and Display, then we'd aggregate it into the NftView.

joshuahannan:
This looks suspiciously like a `{String: AnyStruct}` view. Why do it like this instead of that?

austinkline:
> This looks suspiciously like a `{String: AnyStruct}` view. Why do it like this instead of that?

This is to follow a well-known pattern which is OpenSea's metadata standard, some discussion of that is in #74. A key/value pair isn't necessarily enough, some NFTs on Ethereum also like to specify _how_ the variable should be used which is what the display type is for. It is probably worth checking out their standards page if you haven't before, it can be found [here](https://docs.opensea.io/docs/metadata-standards#attributes)

bjartek:
I have been thinking alot on this Attribute view and personally I think it is the wrong primitive to expose directly. Having code that can transfer the other views into Attributes i see as a way better option. 

My reasoning is that we are basically back to {String:AnyStruct} and that is what we made this standard to avoid. OpenSea does what OpenSea does because they do not have anything that is richer. I think we can do better.

austinkline:
> I have been thinking alot on this Attribute view and personally I think it is the wrong primitive to expose directly. Having code that can transfer the other views into Attributes i see as a way better option.

Translating other views into attributes is not what this is meant to be. This view should yield all the metadata of an NFT, not different formats of that metadata in a non-standard manner. Going the route you propose here would render this view unviable for the two use cases this solves. Those are:

1. Enable creating contract-level features based solely on the metadata of the NFT. A composition of arbitrary views is just not the same here.
2. Platforms showing NFTs are told exactly how to show them. No more special logic for each NFT, and Collection owners themselves have more say in the customization and expression of those items.

> My reasoning is that we are basically back to {String:AnyStruct} and that is what we made this standard to avoid. OpenSea does what OpenSea does because they do not have anything that is richer. I think we can do better.

I would like more reasoning here, please. Why is `{String: AnyStruct}` so bad when you're also proposing yielding arbitrary views instead? If I'm doing a transaction using this view in my contract, I cannot know the underlying type of the view given to me in your proposed option because if I did that I'd need to import every collection I want to support and that just isn't viable, it's why I started discussion on all this in the first place in-fact. Attributes are fundamentally not defined, they must be arbitrary. Every collection can do them differently, and we need a view that can encapsulate that.

If we can do better, I'd like to hear how while also still solving my two use cases because they are both things Flowty needs to be able to address

This view will complete the general things you need for a larger standard NFTView. All dapps would know:
- How to show the nft on a card
- Where the nft is stored
- How to form backlinks for the nft back to its creators' site
- What data exists on the NFT and how to display it/operate on it.

bjartek:
> This view should yield all the metadata of an NFT, not different formats of that metadata in a non-standard manner. 

OpenSea is using a json format because it has no other option, since this is stored in IPFS or behind some other tokenURI scheme. On flow we can use the type system and the richness it provides to make this better. 

The above statments also  suggests that developers duplicate data from the typed views into this untyped view, but that could very well just be made in code. 

> Enable creating contract-level features based solely on the metadata of the NFT. A composition of arbitrary views is just not the same here.

I have a hard problem reasoning about this. In my eyes having typed views are _better_ for discovering features then a loosly typed set of Attributes. You never know if a project uses the same keys or the same types for fields. 

> Platforms showing NFTs are told exactly how to show them. No more special logic for each NFT, and Collection owners themselves have more say in the customization and expression of those items.

Typed views will also tell plattforme exactly how to show things. There is no specific logic for each NFT. I fail to see where this attribute view adds any value on this regard. 

> I would like more reasoning here, please. Why is {String: AnyStruct} so bad when you're also proposing yielding arbitrary views instead? 

Because views have a type, a context and a meaning. You know that if you expose ExternalUrl(url:String) it is an external URL. 

> if I'm doing a transaction using this view in my contract, I cannot know the underlying type of the view given to me in your proposed option because if I did that I'd need to import every collection I want to support and that just isn't viable, it's why I started discussion on all this in the first place in-fact. 

I think we are starting to get into the gist of the problem here. The _intend_ behind the MetadataStandard is that views that are used for interoperability between different consumers and producers are _not_ in their own contracts. It is in a shared place so that people can reason about it. 

I totally agree that there _should_ be some form of Attribute view for things that are not covered in the standard views, but I do not think it should duplicate the data that is already structured and already there. 

> How to show the nft on a card

This is the exact use-case behind the Display view. If you have an arabitrary Attribute view, how do you know what keys to show in a card? Because no keys are mandatory some people can use different keys. And then cards will be displayed differently all over the place. 

> How to form backlinks for the nft back to its creators' site

The new CollectionDetailView have a externalURL link to a collection and the ExternalUrl view have a link back to the actual NFT on another page. This is structured data that you know how to guess. You do not have to guess an Attribute key.

> What data exists on the NFT and how to display it/operate on it.

For arbitraty data that is not part of the other views i totally agree. This is needed.

bluesign:
@austinkline 

I am super confused to be honest, we started with MetadataViews, now we ended up with:

>  Why is {String: AnyStruct} so bad ?

I mean I think we need to be honest to each other, if what you need is  {String: AnyStruct} trying to shape MetadataViews to  {String: AnyStruct} is not the correct approach. Then you should be against MetadataViews from the ground. 

All purpose of metadataviews is to use cadence type system for metadata. 

I understand your need, and support your fight, but the way to do this, is not to make metadataviews an abomination.

austinkline:
Closing this, not going to argue all this in a PR, that isn't what this was supposed to be. @psiemens I will circle back to #74 since clearly things are not as agreed to as I thought.

PR #88, opened 2022-05-18
Adds Edition View for Multiple Editions

joshuahannan:
Adds a view based on the edition view in https://github.com/onflow/flow-nft/pull/75 that accounts for the fact that an NFT might be a part of multiple editions.

```
/// Edition is an optional view for collections that issues multiple objects
    /// with the same or similar metadata, for example an X of 100 set. This information is 
    /// useful for wallets and marketplaes.
    ///
    /// An NFT might be part of multiple editions, which is why the edition information
    /// is returned as an arbitrary sized array
    /// 
    pub struct Edition {

        /// An arbitrary-sized list for any number of editions
        /// that the NFT might be a part of
        pub let infoList: [EditionInfo]

        init(_ infoList: [EditionInfo]) {
            self.infoList = infoList
        }
    }

    /// Edition information for a single edition
    pub struct EditionInfo {

        /// The name of the edition
        /// For example, this could be Set, Play, Series,
        /// or any other way a project could classify its editions
        pub let name: String?

        /// The edition number of the object.
        ///
        /// For an "24 of 100 (#24/100)" item, the number is 24. 
        ///
        pub let number: UInt64

        /// The max edition number of this type of objects.
        /// 
        /// This field should only be provided for limited-editioned objects.
        /// For an "24 of 100 (#24/100)" item, max is 100.
        /// For an item with unlimited edition, max should be set to nil.
        /// 
        pub let max: UInt64?

        init(name: String?, number: UInt64, max: UInt64?) {
            self.name = name
            self.number = number
            self.max = max
        }
    }
 ```

### CODE REVIEW

File: contracts/MetadataViews.cdc [hunk 1]
```diff
@@ -133,6 +133,53 @@ pub contract MetadataViews {
...
+        /// 
+        /// This field should only be provided for limited-editioned objects.
+        /// For an "24 of 100 (#24/100)" item, max is 100.
+        /// For an item with unlimited edition, max should be set to nil.
+        /// 
+        pub let max: UInt64?
+
+        init(name: String?, number: UInt64, max: UInt64?) {
+            self.name = name
+            self.number = number
```
satyamakgec, line 178: Tip: Need to check `number <= max` if max is != nil.

satyamakgec (approved):
LGTM

PR #90, opened 2022-05-27
Use post-conditions and execute phase

joshuahannan:
Makes changes to the transactions:

* Adds some more comments
* Use post phase in some transactions
* use execute phase in some transactions


transactions/destroy_nft.cdc
```diff
@@ -1,14 +1,30 @@
...
 transaction(id: UInt64) {
+
+    /// Reference that will be used for the owner's collection
+    let collectionRef: &ExampleNFT.Collection
+
     prepare(signer: AuthAccount) {
-        let collectionRef = signer.borrow<&ExampleNFT.Collection>(from: ExampleNFT.CollectionStoragePath)
+
+        // borrow a reference to the owner's collection
+        self.collectionRef = signer.borrow<&ExampleNFT.Collection>(from: ExampleNFT.CollectionStoragePath)
```
satyamakgec, line 14: As per the secure cadence - `borrow` is no more returning the `nil` if it doesn't find the type at the given storage. So it should check type before or just remove the panic and allow it to throw the runtime error.

transactions/mint_nft.cdc
```diff
@@ -17,20 +17,38 @@ transaction(
...
 
+    /// Previous NFT ID before the transaction executes
+    let mintingIDBefore: UInt64
 
     prepare(signer: AuthAccount) {
+        self.mintingIDBefore = ExampleNFT.totalSupply
+
         // borrow a reference to the NFTMinter resource in storage
         self.minter = signer.borrow<&ExampleNFT.NFTMinter>(from: ExampleNFT.MinterStoragePath)
             ?? panic("Could not borrow a reference to the NFT minter")
```
satyamakgec, line 34: Same here

satyamakgec (commented):
LGTM, Only a small comment.

transactions/destroy_nft.cdc
joshuahannan: We're using the secure cadence version here so this is still valid syntax. as far as I know, it will also return `nil` if there is nothing in storage, so this is still fine.

satyamakgec: If you say so. I don' have a problem with it but here is a piece of paraphrase from the blog posted by Jan. 

```
The load, copy and borrow functions now perform a force cast
:bulb: Can be identified via analysis tool.

Change description:
load copy and borrow no longer return nil when the type of the value in storage does not match the supplied 
argument. Instead, a runtime error is produced. More information about this change can be found in this FLIP: 
[https://github.com/onflow/flow/pull/718 11](https://github.com/onflow/flow/pull/718)


Update instructions: Developers will need to audit the uses of these functions to ensure they are never called with the
incorrect type. In locations where developers wanted the nil returning behavior if the stored value had the wrong type,
they can use the new type function on AuthAccount to check the type of a value in storage
 ([https://docs.onflow.org/cadence/language/accounts/#account-storage-api 15](https://docs.onflow.org/cadence/language/accounts/#account-storage-api))
[https://github.com/onflow/cadence/pull/1286 8](https://github.com/onflow/cadence/pull/1286)
```

alilloig (approved):
Looks great, I just suggested to change the message on the panic indicating that there is no collection / minter to be borrowed on that path since it seems the only case when it will trigger

PR #92, opened 2022-05-31
added rarity

### LINKED ISSUE #60: Rarity View

bjartek:
When we did the File view we did not give this a name like we have done for Editions/Royalties. So if a project want to export multiple Media files then we need a view that 

closes #60


contracts/MetadataViews.cdc
```diff
@@ -403,4 +403,48 @@ pub contract MetadataViews {
...
+    /// Note that a rarity needs to have either score or description but it can has both
+    pub struct Rarity {
+
+        /// The name of the rarity
+        /// For example, this could be Body,Head
+        /// or any other way a project could classify its editions
+        /// leave it blank to signal that this is the main rarity for all the content of this NFT
+        pub let name: String?
+
+        /// The score of the rarity if this is NFT represents its rarity as a score
```
joshuahannan, line 432: Can you reword this? It is a bit confusing as-is

bjartek:
> This looks pretty similar to what is being discussed about the attributes view. We should make sure that we align on that because I think that would cover the needs that are addressed here. Or am I mistaken?

For me Attributes is key/value/displayHint of things that have no other Typed view that we have a shared understanding about. I could see code that converts Rarity into Attributes if somebody needs it that way. But the ability to have Rarity as its own standalone thing makes it much easier to create conditional logic based on presence/absence of a Rarity.

austinkline:
Pulled these changes into #97 since rarity is embedded into traits. @bjartek I left out the Rarities struct fyi, wasn't sure if it was still needed with how traits will be organized

bjartek:
I 

> Looks good! Can you add the rarity view to the `ExampleNFT` contract also?

I thought that this would be merged into the Trait PR #97 and that this PR will just be closed?

joshuahannan:
ok, I'm fine with closing this since it is in the other PR

Issue #95, opened 2022-06-01
How to use views that are wrappers around other views


bjartek:
## Instructions

Right now we have a pattern where we have Royalties that contains many Royalty. I think the document should be clear on how a producer is supposed to use these. What should you do if you only have a single Royalty. Should you still expose them boxed in Royalties?

The reason I am askin is that there are really several ways of exposing even multiple Royalties today and I think the standard instruction should be clear on how you are supposed to do this:

All of these types are valid forms of Exposing multiple Royalties
 - Royalties
 - [Royalty]
 - {String:Royalty}
 - Royalty //if you onlye have one

If all marketplaces/consumers have to check all this the code gets a little tedious. 

Recommendation:
 - producers should always wrap Royalty in Royalties
 - the same should apply for all other instances of this same pattern. So Editions/Media that is currently merged.

PR #97, opened 2022-06-04
Traits view

austinkline:
Add Traits view to return the traits of an NFT.

You can find discussion about this view in #74 

Some snippets from the original ticket:

Traits - A way to represent the underlying metadata or an item. This would replace views like TopShot's [TopShotMomentMetadataView](https://github.com/dapperlabs/nba-smart-contracts/blob/master/contracts/TopShot.cdc#L462) so that we can standardize what types are being vended from all these different collections set to come out once permission-less deployment is enabled at our June 15th spork.

An Attribute really is a wrapper around a single key-value pair and tells you:

- What is the value?
- What is the name of the value?
- How should I show this value? (Optional)

The last piece (how to show a value) is the key part here for why it is different than another `{String: AnyStruct}` dictionary. Some pieces of metadata are more useful when contextualized with an additional disclaimer.

Take, for example, an attribute that represents the time in which the item was minted. Most likely, that attribute is going to be the block timestamp accessible to us in cadence and will be a number. So, instead of showing a number as the dateMinted attribute, I can set its display type to Date and platforms know to take that number and attempt to parse it differently. So now we've gone from returning

```
{
  "dateMinted": 1653409561
}
```

to

```
[
  {
    "traitType": "dateMinted",
    "value": 1653409561,
    "displayType": "Date"
  }
]
```


contracts/MetadataViews.cdc
```diff
@@ -415,4 +415,55 @@ pub contract MetadataViews {
...
         }
     }
+
+    // An NFT can add MetadataViews.Rarity as a top-level view to define the total rarity.
+    pub struct Rarity {
+        pub let score: UFix64?
+        pub let description: String?
+
+        init(score: UFix64?, description: String?) {
+            self.score = score
```
bjartek, line 425: Should we assert here that you need to atleast specify one?

austinkline: Probably we should sync the rarity struct you defined in #92 here?

austinkline: Sync'd this PR with #92

contracts/MetadataViews.cdc
```diff
@@ -415,4 +415,67 @@ pub contract MetadataViews {
...
+            self.displayType = displayType
+            self.traitType = traitType
+            self.value = value
+            self.rarity = rarity
+        }
+    }
+
+    // helper function to easily convert a dictionary to traits. For NFT collections that do not need either of the
+    // optional values of a Trait, this method should suffice to give them an array of valid traits
+    pub fun dictToTraits(dict: {String: AnyStruct}): [Trait] {
```
joshuahannan, line 449: Should we provide a `{String: String}` dict to traits helper function also? A lot of projects use that

contracts/MetadataViews.cdc
```diff
@@ -415,4 +415,67 @@ pub contract MetadataViews {
...
+            let trait = Trait(traitType: k, value: dict[k]!, rarity: nil, displayType: nil)
+            traits.append(trait)
+        }
+
+        return traits
+    }
+
+    /// Rarity information for a single rarity
+    //
+    /// Note that a rarity needs to have either score or description but it can has both
```
joshuahannan, line 461:
```suggestion
    /// Note that a rarity needs to have either score or description but it can have both
```

joshuahannan (commented):
Looks good so far! Should we also include `Rarity` in the `ExampleNFT` contract resolver, or do we think that it will only be used for traits?

Also, how do we feel about having a `Traits` view that just includes an array of `Trait` like we do for the other views?

contracts/MetadataViews.cdc
austinkline: Hmm would a `{String: String}` not be able to be converted to `{String: AnyStruct}`? Not opposed to more helper functions either way!

austinkline:
> Looks good so far! Should we also include `Rarity` in the `ExampleNFT` contract resolver, or do we think that it will only be used for traits?

Rarity I debated about before submitting this. I wanted to showcase the helper function to generate traits that don't need special handling. I could add a third rarity trait as well if we want and it would just be separate from the other two? 

> Also, how do we feel about having a `Traits` view that just includes an array of `Trait` like we do for the other views?

I was thinking about this, I generally don't like having Object.object kind of syntax, like a `Traits` struct having a `traits` field in it. At the same time, all our other views are non-array types. Sounds like there are other places where this is up in the air as @bjartek has asked about in https://github.com/onflow/flow-nft/issues/95

contracts/MetadataViews.cdc
joshuahannan: You'd probably just have to make all the values strings, even if they represent other types. I think it would still be useful

bjartek:
I think adding Traits makes sense since we already have the pattern in all the other views.

austinkline:
> I think adding Traits makes sense since we already have the pattern in all the other views.

I can live with that, sure. Not my ideal but I agree it's better we adhere to established precedent, especially if it's just nitpicks like this. Will update tonight

aishairzay (approved):

contracts/MetadataViews.cdc
```diff
@@ -415,4 +415,96 @@ pub contract MetadataViews {
...
+        return Traits(traits)
+    }
+
+    /// Rarity information for a single rarity
+    //
+    /// Note that a rarity needs to have either score or description but it can have both
+    pub struct Rarity {
+        /// The score of the rarity as a number
+        ///
+        pub let score: UFix64?
```
psiemens, line 494: Should the `Rarity` struct make any attempt to standardize a score's scale?

e.g. I could imagine something like `maxScore = 100` so that score is rendered as `42/100`

psiemens (commented):
Looks good overall! Left a couple suggestions and then one question about the rarity score field.

contracts/MetadataViews.cdc
bjartek: I think it would be wise to add something like that. That greater should be more rare and the max should be 100.0

contracts/MetadataViews.cdc
```diff
@@ -415,4 +415,96 @@ pub contract MetadataViews {
             self.socials = socials
         }
     }
+
+    // A view to represent a single field of metadata on an NFT.
+    //
+    // This is used to get traits of individual key/value pairs along with some contextualized data about the trait
+    pub struct Trait {
+        // The name of the trait. Like Background, Eyes, Hair, etc.
+        pub let traitType: String
```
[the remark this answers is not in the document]
austinkline: Not opposed, `traitType` was what I used purely because it's what OpenSea calls it, no strong opinions there and `name` would probably be more easily understood/intuitive

psiemens (approved):
LGTM 👍

aishairzay (approved):
LGTM !

joshuahannan:
I still think it would be good to have a `stringDictToTraits(_ dict: {String: String}): Traits` utility function. Does anyone object to adding that? Also have you run all the tests with `make test` to make sure everything passes? I still haven't figured out how to get CI to run for PRs from forks

austinkline:
> I still think it would be good to have a `stringDictToTraits(_ dict: {String: String}): Traits` utility function. Does anyone object to adding that? Also have you run all the tests with `make test` to make sure everything passes? I still haven't figured out how to get CI to run for PRs from forks

So just to make sure I understand the value-add, shouldn't I be able to feed in a {String: String} to something with {String: AnyStruct}?

All tests pass using `make ci`

joshuahannan (approved):
You're right, I guess it does work. I didn't know if the type system would be ok with that. I'm fine with it then

PR #100, opened 2022-06-16
Helpers

bjartek:
The Metadata standard is very generic in its nature and I have found several error cases that I think warrants helpers to resolve the core views. 

One liner solutions can panic your entire system if an NFT say they resolve Display but return just a String. And since the interface is to return AnyStruct? they will get no compiler erorr when doing so.


bshahid331:
+1 For this, this is something the catalog could leverage as well to extra views more safely. Changes LGTM.

contracts/MetadataViews.cdc
```diff
@@ -70,6 +70,16 @@ pub contract MetadataViews {
         }
     }
 
+    ///a helper to get Display in a typesafe way
```
joshuahannan, line 73:
```suggestion
    /// A helper to get Display in a typesafe way
```

Can you do this to all the comments?

joshuahannan (commented):
Looks good! Just had one comment. Also, can you write some scripts/tests for these and run everything with `make test`?

contracts/MetadataViews.cdc
bjartek: Sure.

bjartek:
> Looks good! Just had one comment. Also, can you write some scripts/tests for these and run everything with `make test`?

Would it be possible to do this with overflow? Are the generated files used anyhwere else?

joshuahannan (approved):
Thanks for updating the scripts. I want to integrate overflow, but we still need something for this I think: https://github.com/bjartek/overflow/issues/9

I'll see what I can do

joshuahannan:
It looks like there are some merge conflicts also

bjartek:
> Thanks for updating the scripts. I want to integrate overflow, but we still need something for this I think: [bjartek/overflow#9](https://github.com/bjartek/overflow/issues/9)
> 
> I'll see what I can do

We kind need to do a pr to flowkit Deploy code then to be able to specify a base path or something I think.

PR #101, opened 2022-06-16
fix build errors

austinkline:
When I accepted some commits on #97 (comment misspellings/rewording) it looks like it broke part of the `go tidy` check, this fixes that. (sorry!)


austinkline:
@joshuahannan looks like test runs can be approved by repo owners:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #102, opened 2022-06-23
Organize content for docs refresh caching

codingone21:
Closes https://github.com/onflow/next-docs-v1/issues/201

- Add workflow for refreshing docs cache with a docs/ content change
- Copy readme into the /docs content folder

### DIFF
--- a/.github/workflows/call-refresh-doc.yml
+++ b/.github/workflows/call-refresh-doc.yml
@@ -1,0 +1,51 @@
+# This workflow is stored and called from each repository that contains documentation files
+name: call-refresh-doc
+on:
+  push:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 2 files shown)


joshuahannan (approved):
So now when we update the README, will we also have to update this `overview` doc? Or will it auto-update?

codingone21:
> So now when we update the README, will we also have to update this `overview` doc? Or will it auto-update?

Unfortunately, we will need to update both :( The idea is to create a consistent directory structure repo/docs/page.md for any markdown files that should be present on the documentation site! Technically, they don't need to be exactly the same, but as of now, the only file that we are pulling for the docs is the README page.

codingone21:
Thanks for the review! 👍

Issue #98, opened 2022-06-04
Core Views in NFTView


austinkline:
**This is a spin-off ticket originating from #74. Please review its comment history for deeper context about what this ticket is for**

A central theme of discussion many of us have brought up is what views should be "required" to be implemented in order to say that a collection has implemented Flow's metadata standard. While individual collections can define their own views, there is also a central set of what some of us have called "core" views 

## What are core views for?

Core views, when all put together, should give a "complete" picture of an NFT on flow. That is.. marketplaces, wallets, and applications should be able to use the NFTView as a central source to obtain information about an item both to transact with and to display. 

## Current Proposed Core Views
- [Display](https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L39) - "How do I show this NFT?"
- [ExternalURL](https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L312) - "Where did this NFT come from/how do I make backlinks to this NFT's project?"
- [Medias](https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L273) - "What is the full image/animation of this NFT? Are there multiple ways to show it?"
- Traits (In PR at #97) - "What data exists on this NFT? What is its Body, Head, Background, Hometown, etc? How do I show this data?"

## Possible missing pieces
- [NftCollectionData](https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L323) - "How should I store this NFT?"
- TokenURI - Some collections like TopShot have data that exists off-chain. Ideally we can mark how to obtain that extra data or at least note that what an application has could be incomplete

I'm not sure about NftCollectionData, I've flip-flopped on it a few times. TokenURI I think has its place given that some prominent projects are already split, so we know that we will need to handle that.

If anything else is missing, please chime in! The ideal is that we settle on these core views and then we can create a factory method or some kind of helper on the MetadataStandards contract that builds this for you. Then, a collection can just delegate to the factory for building the NftView and it can be practically for free for collection owners to include.

austinkline:
cc @psiemens @bjartek @bluesign

bjartek:
NFTCollectionView and Editions?

austinkline:
> NFTCollectionView and Editions?

Editions I'm unsure of. Does it belong on the top-level? Or should we instead encourage collection owners to make that into a trait?

spacepluk:
[...]

PR #103, opened 2022-06-28
Add core nft view

bshahid331:
See Issue: https://github.com/onflow/flow-nft/issues/98

Implementing the core NFT View which consists of common views that NFTs should implement to have basic interoperability throughout the ecosystem.


contracts/MetadataViews.cdc
```diff
@@ -33,6 +33,36 @@ pub contract MetadataViews {
         pub fun getIDs(): [UInt64]
     }
 
+    /// NFTView is a group of views used to give a complete picture of an NFT
+    ///
+    pub struct NFTView {
+        pub let id: UInt64
```
austinkline, line 39: we might also want uuid in here

transactions/scripts/get_nft_view.cdc
```diff
@@ -0,0 +1,111 @@
+import ExampleNFT from "../../contracts/ExampleNFT.cdc"
+import MetadataViews from "../../contracts/MetadataViews.cdc"
+
+pub struct NFTView {
```
joshuahannan, line 4: Is is necessary to define a new NFTView in here? Can't we just return the one from the view?

joshuahannan: ah, that makes sense

PR #107, opened 2022-07-13
recepient -> receiver,

bjartek:
recepient is spelled recipient, but I belive receiver is a batter label here. 

I noticed this because never version of cadence actually error if you use the wrong labels!


bluesign:
I think correct one is "royalty recipient" vs "royalty receiver" , but English is not my first language, so I can be wrong.

joshuahannan:
This would be a breaking change, so I don't think it is worth changing.

bjartek:
Actually right now cadence does not care about the labels of structs created in other contracts, i found out this yesterday. But I am ok if we want to have a spelling error in the contracts forever :)

bluesign:
I agree that this is breaking change, with current handling of breaking changes not worth it, but on the other hand I think we care more than necessary about breaking changes. I mean running after dapps to make them fix contracts is a losing battle.

joshuahannan:
@bjartek So you're saying that if we did this upgrade now, it would not be breaking, but when that version of cadence is applied, then it would break the contracts that use this? That would be a breaking change anyway, so maybe this is still worth merging and updating

bjartek:
Yes.

bjartek:
merged upstream master.

joshuahannan:
@bjartek you'll probably also need to run `make test` to build the assets so we can merge

bjartek:
done

joshuahannan:
The back and forth on this was taking too long, so I just created a PR myself. I'll close this

Issue #120, opened 2022-08-29
Reformat NonFungibleToken Go tests in JavaScript


sisyphusSmiling:
### Issue To Be Solved
In keeping with the tests to be written for NFTForwarding contract (See issue https://github.com/onflow/flow-nft/issues/10 and PR https://github.com/onflow/flow-nft/pull/116), tests for the NonFungibleToken contract should be written in JavaScript, leveraging `flow-js-testing`.

austinkline:
fyi there has also been discussion about moving this repo to use overflow:
https://github.com/onflow/flow-nft/issues/105

And there is already some example usage for how to make that happen from @bjartek 
https://github.com/bjartek/flow-nft-overflow

joshuahannan:
https://github.com/onflow/flow-nft/pull/116 Addresses this, so we can close this when that gets merged

sisyphusSmiling:
Closed by PR#121

PR #116, opened 2022-08-18
NFTForwarding Contract + Introduce JS Test Suite

sisyphusSmiling:
Introduces NFTForwarding contract containing an NFTForwarding resource implementing NonFungibleToken.Receiver which allows forwarding of deposited NFTs to a designated recipient's NonFungibleToken.CollectionPublic. PR closes issue https://github.com/onflow/flow-nft/issues/10

In addition to the introduction of NFTForwarding account, this PR also includes a JavaScript test suite leveraging flow-js-testing. Tests are found in lib/js/test, adopting the existing directory structure found in lib/go with test helper methods and const values found in lib/js/templates.

Tests associated with this contract demanded new transactions to cover all test cases. These new transactions include:

* `create_forwarder.cdc` - creates NFTForwarder resource with passed Address as recipient
* `transer_nft_to_receiver.cdc` - similar to transfer_nft.cdc, but deposits NFT to NonFungibleToken.Receiver instead of NonFungibleToken.CollectionPublic
* `unlink_collection.cdc` - unlinks the signer's ExampleNFT.ExampleNFTCollectionPublic from ExampleNFT.CollectionPublicPath
* `unlink_forwarder_link_collection.cdc` - unlinks the signer's NFTForwarder in public storage and link's the signer's ExampleNFT.ExampleNFTCollectionPublic, replacing the public forwarding capability with a collection capability.

To support JS testing suite, GitHub CI workflow was amended to include node and Flow CLI dependency setup. These changes were largely modeled on the ci workflow outline found in the `onflow/flow-ft` repo .

PR #121, opened 2022-08-30
Refactored NonFungibleToken Go tests to JavaScript. Repo now fully ported to both test suites with matching coverage.

sisyphusSmiling:
Issue addresses #120 

This PR focuses on porting over Go tests cases covering the NonFungibleToken contract to JavaScript using `flow-js-testing`. This is a follow up to PR #116 which introduced the NFTForwarding contract and the JavaScript test suite to this repo.

Most test cases follow the logic of those written in Go, with the goal of matching or increasing coverage in the refactor. To support this, additional templates were added (to make tests cleaner & easier to read) and a new script. This new script - `get_collection_ids.cdc` - simply returns an array `[UInt64]` of the NFT IDs in the passed address's collection.


lib/js/test/tests/nft_forwarding_test.test.js
```diff
@@ -145,12 +145,12 @@ describe("NFTForwarding Contract Tests", () => {
...
 
         // Unlink collection in recipientAccount
-        const [unlinkTxn, e1] = await shallPass(
+        await shallPass(
             sendTransaction("unlink_collection", [recipientAccount], [ ])
         );
 
         // Transfer NFT, but attempt to forward should fail
-        const [transferTxn, e2] = await shallRevert(
+        await shallRevert(
```
satyamakgec, line 153: I prefer to make sure the revert string is not null as there are some cases where revert can happen because of some unexpected things in code execution. The current way it is not making sure that it is failing at the same desired statement.

scripts/get_collection_ids.cdc
```diff
@@ -0,0 +1,15 @@
+import NonFungibleToken from "../../contracts/NonFungibleToken.cdc"
+import ExampleNFT from "../../contracts/ExampleNFT.cdc"
```
satyamakgec, line 2: I am thinking to make the scripts and transactions are generic and only based on the interfaces, If possible. Here we can provide the `CollectionPublicPath` as the parameter like a string and typecast to the publicPath is possible. 

Although I am not rigid about this just want to know the thoughts of the team. It is better to make it generalise so the other developers wouldn't need to change anything init.

sisyphusSmiling: I can see pros and cons to either approach. Having generalized scripts in our repo that others can use across different implementations of `NonFungibleToken` would be great. The benefit of generalizability is paid for by passing the path. I think that's probably worth the benefit, especially if the path is a stored as a variable in the implementation contract as it's easily referenced.

satyamakgec: Yeah so do you mind updating the script to support the universal NFTs

lib/js/test/tests/nft_forwarding_test.test.js
satyamakgec: any input here @sisyphusSmiling?

sisyphusSmiling: I agree. Submitted changes in the recent commits

scripts/get_collection_ids.cdc
sisyphusSmiling: Made sure to add change to recent commits ✅

satyamakgec (approved):
Looks good to me.

scripts/get_collection_ids.cdc
psiemens: Did you mean to remove this import? I noticed that the script no longer requires `ExampleNFT`.

Also, I think the generalized version is better as well!

lib/js/test/templates/assertion_templates.js
```diff
@@ -0,0 +1,47 @@
...
+export async function assertTotalSupply(expectedTotalSupply) {
+    const [actualTotalSupply, e] = await executeScript(
+        "get_total_supply"
+    );
+    assertNoError(e);
+    expect(actualTotalSupply).toBe(expectedTotalSupply.toString());
+};
+
+// Asserts whether length of account's collection matches
+// the expected collection length
```
psiemens, line 25: This comment is copied from `assertCollectionLength` above

lib/js/test/templates/assertion_templates.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+};
+
+// Simple function that asserts passed condition is satisfied
+// by given expression
+export function assert(condition, expression) {
+    if (expression != condition) {
+        throw Error("Assertion failed");
+    };
+};
```
psiemens, line 47: Why not just use Jest here?

```js
expect(condition).toEqual(expression)

File: lib/js/test/templates/assertion_templates.js [hunk 5]
```diff
@@ -0,0 +1,47 @@
...
+
+// Asserts that passed error is null
+// Exits emulator on failure
+export function assertNoError(error) {
+    if (error != null) {
+        console.log(`Error in deployment - ${error}`);
+        emulator.stop();
+        process.exit(1);
+    };
+};
```
psiemens, line 39: Could this be a custom Jest matcher? https://jestjs.io/docs/expect#expectextendmatchers

I'm not sure if this makes sense, but maybe something like this:
```js
expect(error).noEmulatorError()

psiemens: Maybe a custom matcher doesn't make sense, but I think it's good to always report failures to Jest and let it exit the process if needed, rather than bypassing its assertion framework.

psiemens (commented):
Nice! I love that we're making use of the JS testing library here.

alilloig (approved):
LGTM!!

joshuahannan (approved):

joshuahannan:
@sisyphusSmiling You'll need to merge the latest changes from master and push it up again to resolve the conflicts

Issue #123, opened 2022-09-01
Adding Medias view to MetadataViews.NFTView


bymi15:
## Issue To Be Solved

Are there any reasons why the `Medias` view is not included in the `MetadataViews.NFTView`?
Can this be included as there are quite a few NFT collections that use video files rather than images and this can only be included in the `Medias` view.

bjartek:
We cannot add state to this struct as it is already created, then you would need to have a new version of the view with a new name.

bymi15:
I’m guessing that’s due to the limitation of flow contract updatability.
In that case, NFTViewV2 sounds like the way to go. 
Maybe there should be some further discussions on other areas of improvements before deciding what needs to be added to V2.

joshuahannan:
@bymi15 Can you open a PR with this update? That would be a good place to have this discussion

bymi15:
> @bymi15 Can you open a PR with this update? That would be a good place to have this discussion

Thanks for your response @joshuahannan
I have just raised a PR as you've suggested.

PeterBenc:
IMO this is quite an important issue, as far I understand it, the only source of `Media` for a single nft is the `thumbnail` field which isn't of type `Media` (honestly its pretty strange that colleciton-specific data had media files but nfts don't). I was hoping we could migrate from alchemy to nft catalog but since all I have is `a thumbnail` of which I don't even know what `mimeType` it is I really can't :( 

Pls correct me if I am wrong but seems like as long nft media goes, the `thumbnail` field is the best I can do right now.

bymi15:
@PeterBenc I think the `thumbnail` field was intended more for small-medium sized images that can be displayed on websites.
Totally agree that there should be a field that allows one or more media files (i.e. the `medias` view should be included in the core NFTView)

joshuahannan:
As we said in the PR, the NFTView is not a requirement, it is simply a utility view that combines all other views. If you want `Medias` to be included in a whole view, you can simply define a view in your script that combines `Medias` with `NFTView` and return that. I don't think we're at the point yet where we want to define a new NFTView. I'm tempted to just close this issue

bymi15:
The problem is that this is all tied in with NFTCatalog which uses NFTView: https://github.com/dapperlabs/nft-catalog/blob/main/cadence/contracts/NFTRetrieval.cdc#L99

Perhaps this could be raised in NFTCatalog - so they can support other metadata views such as Medias, then we could probably close this one.

[...]

PR #127, opened 2022-09-14
WIP: Add NFTViewV2 to MetadataViews contract

bymi15:
Closes: #123

## Description

- Add a new NFTViewV2 struct in order to include additional optional fields
- Add a 'medias' field to NFTViewV2

### DIFF
--- a/contracts/MetadataViews.cdc
+++ b/contracts/MetadataViews.cdc
@@ -88,6 +88,68 @@ pub contract MetadataViews {
         )
     }
 
+    /// NFTViewV2 is an enhancement of NFTView with additional information
+    /// including the medias view as an optional field
+    ///
+    pub struct NFTViewV2 {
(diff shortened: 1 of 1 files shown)


bymi15:
I haven't added any tests yet as this is still a draft PR.
There could be other fields we want to add to NFTViewV2.

bymi15:
> I'm starting to think that maybe it isn't a good idea at this point to introduce a new NFTView. I think it might be better for now to just define this view within a script, and have it contain the normal NFTView and the medias view and just return that. I don't want to add second versions of things too prematurely in case there are other things in the future people want to add.

Yeah I completely agree, could close this PR but leave the issue open so it could be considered in the future when there are enough changes that would make sense to introduce a V2.

joshuahannan:
good suggestion. I'll close this

Issue #125, opened 2022-09-08
Add `containsNFT` method to CollectionPublic after default implementation is in place


bjartek:
Right now if you want to check if an NFT is in a collection and _not_ panic if it fails you have to get all ids and then check if your id is in that array. With the way that storage is currently implemented this is expensive and inefficient.

This proposal suggests adding a `containsNFT` method that will return a boolean to signal if an NFT with a given id is here or not. 

Note that simply adding this change to the chain might not be enough. Because if the computation of said method will increase as the collection grow it is not a sustainable solution. This issue also touches on the subject https://github.com/onflow/cadence/issues/1938

bluesign:
Very good idea, also we can add something like `borrowNFTSafe(id: UInt64): &NFT?` for convenience. 

I think computation there will be not much problem, checking should be `O(logN)`

bjartek:
I almost think that when default implementation is in we should make nft 1.1 that is bc with NFT with just default impl. It is going to make life so much easier.

joshuahannan:
I think this is a great idea. We'll make a PR to add this

bjartek:
Do we have an ETA for when default implementations will be added to cadence on testnet/mainnet. Default adding of getViews/resolveViews to NFT resource would also be lovely.

joshuahannan:
I think it is going to be added in the next spork, so we'll get them integrated into the standards soon in preparation

alilloig:
@bjartek do you think there is a way of introducing default implementations for getViews/resolveViews to collection (and to borrowNFTSafe) without breaking the existing implementations of `Collection`? I believe that adding an additional conformance to `MetadataViews.Resolver` to the resource will break those.

bjartek:
getViews/resolveViews are not on the collection, they are on the resource are they not? if we specify that the nft interface needs it and default implement it as empty it should work should it not? So add it as methods on INFT and default implement it there?

joshuahannan:
@bjartek I see, so you're saying that we don't add the `MetadataViews.Resolver` conformance to `NFT` or the `ResolverCollection` conformance to `Collection`, we just add it to `INFT` and add a default implementation there. I think that should work

bjartek:
I think so too, but after default implementations are here, we can just try and see that it works.

PR #134, opened 2022-10-21
add default implementations for metadata views and safeBorrow

joshuahannan:
Closes: #125 

## Description

Adds default implementations to the INFT interface for NFT metadata views
Adds a default implementation of borrowNFT that returns an optional instead of panicing

______

For contributor use:

- [x] Update the version in package.json when there is a change in the smart contracts


bjartek (approved):
Much needed feature!

contracts/NonFungibleToken.cdc
```diff
@@ -71,6 +71,24 @@ pub contract interface NonFungibleToken {
     pub resource interface INFT {
         // The unique ID that each NFT has
         pub let id: UInt64
+
+        /// Function that returns all the Metadata Views implemented by a Non Fungible Token
```
turbolent, line 75: What is the reason for declaring these metadata function in `INFT`? 
Shouldn't the functions in `MetadataViews.Resolver` gain default implementations?

turbolent (commented):
Nice!

contracts/NonFungibleToken.cdc
joshuahannan: because we want to upgrade the NFT standard without breaking every NFT contract. If we require every NFT contract to upgrade to use `Resolver`, then they'll all break

turbolent: I see. Basically the functions are only here to keep backwards-compatibility? 

Maybe add a comment explaining this. New readers of the contract might wonder and have the same question.

joshuahannan: Added a comment. 👍

satyamakgec (approved):
LGTM

contracts/NonFungibleToken.cdc
alilloig: Thank you @turbolent. I have been dealing with the same issue on the FT repo and the best I can say now is that I guess I haven't interiorised yet default function implementations.

alilloig (approved):
LGTM!

austinkline:
@joshuahannan  Any ETA on this being released? It would make getting metadata much easier 🙂

joshuahannan:
Need to do some more testing and coordinate a deployment to testnet for sometime early next week, then we'll probably deploy to mainnet about a week later

joshuahannan:
Finally got it merged! We'll be performing the upgrade on testnet tomorrow at 10:30am PST and then will likely upgrade mainnet a week later

bluesign:
@joshuahannan there is a need for post condition for borrowNFTSafe I guess.  ( sorry I missed this before merge )

joshuahannan:
oh yeah, thank you for pointing that out! I'll add it in another PR, you mean this one, right?
```
            post {
                result == nil || result!.id = id: "The returned reference's ID does not match the requested ID"
            }
```

PR #142, opened 2022-11-30
add post condition to borrowNFTSafe

joshuahannan:
## Description

Adds a post condition to the `borrowNFTSafe` method to check if the reference has the correct ID

______

For contributor use:


alilloig (approved):
Good catch @bluesign!!

sisyphusSmiling (approved):

PR #148, opened 2023-01-30
add a metadata view to designate an nft as a pack

austinkline:
Closes: #???

## Description

Some NFTs on Flow are Packs. They can be opened to obtain the contents inside of them, but are largely useless afterwards. Two examples of this are NFL All Day packs, and Gaia NFT Packs. In both of these cases, there isn't a standard metadata view to designate them as opened. Because of that, users are at risk of purchasing a pack they might not know is empty.

This view adds a way to detect if something is a pack, letting marketplaces decide what they should do with that information. Perhaps that just means displaying that the pack is opened, or it could mean disabling trading altogether. 

______

For contributor use:

### DIFF
--- a/contracts/MetadataViews.cdc
+++ b/contracts/MetadataViews.cdc
@@ -695,6 +695,28 @@ pub contract MetadataViews {
         }
     }
 
+    // Taken from GaiaPackNFT https://flow-view-source.com/mainnet/account/0xfdae91e14e960079/contract/GaiaPackNFT
+    // AllDay Packs have a subset of these so it should be safe to use this more verbose set
+    pub enum PackStatus: UInt8 {
+        pub case Sealed
+        pub case RevealRequested
+        pub case Revealed
+        pub case OpenRequested
+        pub case Opened
+    }
+
+    pub struct Pack {
+        pub let status: PackStatus
+
+        pub fun isOpen(): Bool {
+            return self.status == PackStatus.Opened
+        }
+
+        init(_ status: PackStatus) {
+            self.status = status
+        }
+    }
+
     /// Helper to get Traits view in a typesafe way
     ///
     /// @param viewResolver: A reference to the resolver resource
--- a/docs/MetdataViews/MetadataViews_Pack.md
+++ b/docs/MetdataViews/MetadataViews_Pack.md
@@ -1,0 +1,32 @@
+# Struct `Pack`
+
+```cadence
+pub struct Pack {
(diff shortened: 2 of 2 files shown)


contracts/MetadataViews.cdc
```diff
@@ -695,6 +695,28 @@ pub contract MetadataViews {
...
 
+    // Taken from GaiaPackNFT https://flow-view-source.com/mainnet/account/0xfdae91e14e960079/contract/GaiaPackNFT
+    // AllDay Packs have a subset of these so it should be safe to use this more verbose set
+    pub enum PackStatus: UInt8 {
+        pub case Sealed
+        pub case RevealRequested
+        pub case Revealed
+        pub case OpenRequested
+        pub case Opened
+    }
```
austinkline, line 706: I'm not sure if an enum is the right call given its finality. We could instead make it a string so we can add to this as needed. Both Gaia and AD packs use an enum, though

bjartek:
+1

contracts/MetadataViews.cdc
joshuahannan: Could we potentially make it an integer, and then have a switch statement to indicate what its status is? that way we can add more to it if needed, but it is still similar to an enum

joshuahannan (approved):
I think this looks pretty good! I'll try to get some of the studio engineers to look at it also

contracts/MetadataViews.cdc
austinkline: No strong feelings on that part either way. I was talking to Chris on the Gaia team and he had suggested maybe going with a "Consumable" standard instead where packs are just a one-time consumable. In which case, we would amend this to let you ask how many charges are left in some capacity.

I'll open a ticket to discuss this and then we can amend this PR as needed, it is just a starting point for discussion 🙂

austinkline: https://github.com/onflow/flow-nft/issues/149

austinkline:
Added a ticket so we can discuss:
https://github.com/onflow/flow-nft/issues/149

bjartek:
The semantics of packs is different in every implementation i have seen. Can we open up a bigger discussion around consumable/pack semantics?

I tried to create a tax reporting export to koinly and some packs/opening logic does not have enough data to cost-allocate the items in the pack/consumable to the new nfts. This is a pretty huge problem if the tax government comes calling.

Issue #146, opened 2023-01-06
Adding support in MetadataViews to support viewing nested resources


benbackend:
### Issue To Be Solved

MetadataViews should be implemented in almost all resources to tell what the resource is about. 
At the moment, resources that are nested has no generic way to expose their views. Projects do it in there own way which is not really useful or efficient.

The proposal here introduce a new function that treats every nesting resource as a "collection" of the nested resources. By implementing MetadataViews.ResolverCollection to the "collection" resource, we can easily get the nested resources and it's resolver(i.e. the views).

### Suggest A Solution

Additions to MetadataViews.ResolverCollection interface 
```cadence 
    pub resource interface ResolverCollection {
        pub fun borrowViewResolver(id: UInt64): &{Resolver}
        pub fun getIDs(): [UInt64]

       // Adding a new function with default implementation here to expose subCollection ResolverCollection 
       // If a resource contains any other resources that implements MetadataViews.Resolver, this should return the resolver of the nested resource
       // return nil by default , which means that the resource does not contain nested resources / querying id does not exist. 
       pub fun borrowSubCollection(id: UInt64): &{ResolverCollection}? {
              return nil 
       }
    }
``` 

An proposed implementation can be seen here: 
https://github.com/crash13override/flovatar/pull/10
Taking Flovatar and FlovatarMarketplace as an example, trying to : 
1. expose Flovatar Components from Flovatar NFT 
2. expose Flovatar NFT from Flovatar SaleItem Collection 

This supports looping quite well.
If you would like to view on Flovatar Components equipped on a Flovatar that is listed : 
Flovatar SaleItem Collection  -> Flovatar NFT -> Flovatar Components

all credits to @bjartek

bluesign:
why not `borrowResolverCollection` on `Resolver` ?

bjartek:
`borrowSubCollection` on `Resolver` could also work here. 

Not really sure how to overcome that limitation. We have hit that already since we have a use case where an NFT can hold two different types of subNFTs. However since they both/all use uuid for ids there is no conflict.

joshuahannan:
@Bamthelearner I think this is a good idea! Can you make a PR for this and integrate the other suggestions in this thread? Thank you!

alilloig:
I like it better as a method of `Resolver`, that way resources that aren't an actual collection could keep implementing that interface rather than `ResolverCollection`.

[...]

PR #155, opened 2023-02-08
Borrow sub collection

bjartek:
Closes: #146 

I added methods to NFT aswell, since there should be a way to reason about this if you are comming to this from a Resolver standpoint and from an NFT standpoint.


contracts/MetadataViews.cdc
```diff
@@ -20,6 +20,11 @@ pub contract MetadataViews {
     pub resource interface Resolver {
         pub fun getViews(): [Type]
         pub fun resolveView(_ view: Type): AnyStruct?
+
+        /// Borrow the subcollection if this resolver has any sub nfts in it
+        pub fun borrowSubCollection() : &{ResolverCollection}? {
+            return nil
+        }
```
joshuahannan, line 27: What if there are multiple collections?

contracts/NonFungibleToken.cdc
```diff
@@ -93,6 +93,24 @@ pub contract interface NonFungibleToken {
...
+
+        /// Returns a borrowed reference to an child NFT of this NFT
+        /// so that the caller can read data and call methods from it
+        ///
+        /// @param id: The ID of the NFT that want to be borrowed
+        /// @return A reference to the NFT
+        ///
+        pub fun borrowChildNFT(id: UInt64): &NFT? {
+            return nil
+        }
```
joshuahannan, line 113: I'm a little confused about these. First, don't we want to also have the `borrowSubCollection` method here?
Second, what is a child NFT in this context?

bjartek: We cannot borrowSubCollection as &{ResolverCollection}here since it would be an import loop. Also NFT cannot implement MetadataViews.Resolver because of the same import loop.

We can always change the child term to use something else, `subNFT` or embedded or another word.

contracts/MetadataViews.cdc
bjartek: Good question, should this method have a classifier? if it does then we need to have another method to actually get what subCollections you have.

contracts/NonFungibleToken.cdc
bjartek: The decition to put NFTCollectionData in MetadataViews and not in here is the cause of this :/

bjartek: I actually think that having subCollection in &{Resolver}and these methods here is confusing, should we not use `borrowSubCollection` in &{Resolver}and standardize on names here?

bluesign:
I feel borrowChildNFT should be on another interface, common usage is not having child NFT afaik. 

But also as this will be deprecated soon, it is not that big of a problem.

contracts/MetadataViews.cdc
joshuahannan: yeah, I don't know if we can standardize how to retrieve a specific collection. There could be different collection types stored, different collections of the same type differentiated by some random piece of metadata, or any other method.

It makes me think we should just have this return everything. Maybe we just return an array of collection references? or a dictionary that maps `{AnyStruct: &{ResolverCollection}?}` so the implementations can decide how they want to key them. Doesn't seem great either. :(

contracts/NonFungibleToken.cdc
joshuahannan: Can't we still have `borrowSubCollection` here and in metadata views and just don't import metadata views like we do with the other resolver methods? Maybe I'm misunderstanding something

bjartek: What is borrowSubCollection here going to return then?

bjartek: One problem here is that there is no interface that covers the case of read access to all sub nfts. CollectionPublic also has deposit.

joshuahannan: It could just return an optional

joshuahannan:
@bluesign What will be deprecated soon?

bluesign:
> @bluesign What will be deprecated soon?

My bad, I thought this will be in the collection somehow. Then I think `borrowChildNFT` should be separate interface for sure, I don't think many NFTs have/will have child NFTs.

joshuahannan:
This is handled in the v2, standard, right? Can we close this?

PR #161, opened 2023-03-03
Contract resolver

austinkline:
## Description

Adds a new contract interface called Resolver which can be implemented and borrowed to obtain metadata about a contract.

This will help platforms with discoverability, a contract could expose the metadata view `NFTCollectionData` on the contract itself now instead of a specific NFT so that you no longer need to be able to borrow an NFT to get important setup data for a collection
______

For contributor use:

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,35 @@ to your chosen fungible token for now. Then, use that public path for your royal
 and in the future, you will be able to easily update the link at that path to use the
 fungible token switchboard instead.
 
+## Contract metadata
+
+Now that contract borrowing is released, you can also implement the [Resolver](./contracts/Resolver.cdc) interface on your contract
+and resolve views from there. As an example, you might want to allow your contract to resolve NFTCollectionData and NFTCollectionDisplay
+so that platforms do not need to find an NFT that belongs to your contract to get information about how to set up or show your collection
+
+```cadence
+import Resolver from 0xf8d6e0586b0a20c7
+import MetadataViews from 0xf8d6e0586b0a20c7
+
+pub fun main(addr: Address, name: String): AnyStruct? {
+  let t = Type<MetadataViews.NFTCollectionData>()
+  let borrowedContract = getAccount(addr).contracts.borrow<&Resolver>(name: name) ?? panic("contract could not be borrowed")
+
+  let view = borrowedContract.resolveView(t)
+  if view == nil {
+    return nil
+  }
+
+  let cd = view! as! MetadataViews.NFTCollectionData
+  return cd.storagePath
+}
+```
+
+Will Return
+```cadence
+{"domain":"storage","identifier":"exampleNFTCollection"}
+```
+
 ## How to propose a new view
 
 Please open a pull request to propose a new metadata view or changes to an existing view.
--- a/contracts/ExampleNFT.cdc
+++ b/contracts/ExampleNFT.cdc
@@ -11,8 +11,9 @@
 
 import NonFungibleToken from "./NonFungibleToken.cdc"
 import MetadataViews from "./MetadataViews.cdc"
+import Resolver from "./Resolver.cdc"
 
-pub contract ExampleNFT: NonFungibleToken {
+pub contract ExampleNFT: NonFungibleToken, Resolver {
 
     /// Total supply of ExampleNFTs in existence
     pub var totalSupply: UInt64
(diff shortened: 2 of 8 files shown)


austinkline:
Fixed all failing tests! Just a callout that running `make ci` from my own machine led to lots of failures, though. Is something missing there? This same error message (but a new port each time) shows up for lots of tests
```
  ● NFTForwarding Contract Tests › Setup NFTForwarder then unlink and restore collection in forwarder's account

    HTTP Request Error:
          HTTP Request Error: An error occurred when interacting with the Access API.
          error=request to http://localhost:62360/v1/blocks?height=final&expand=payload failed, reason: connect ECONNREFUSED 127.0.0.1:62360
          hostname=http://localhost:62360
          path=/v1/blocks?height=final&expand=payload
          method=GET
```

README.md
```diff
@@ -328,6 +328,35 @@ to your chosen fungible token for now. Then, use that public path for your royal
 and in the future, you will be able to easily update the link at that path to use the
 fungible token switchboard instead.
 
+## Contract metadata
+
+Now that contract borrowing is released, you can also implement the [Resolver](./contracts/Resolver.cdc) interface on your contract
+and resolve views from there. As an example, you might want to allow your contract to resolve NFTCollectionData and NFTCollectionDisplay
+so that platforms do not need to find an NFT that belongs to your contract to get information about how to set up or show your collection
```
joshuahannan, line 335:
```suggestion
so that platforms do not need to find an NFT that belongs to your contract to get information about how to set up or show your collection.
```

joshuahannan (commented):
Thank you so much for adding this! My main suggestion is to choose a more specific name than `Resolver`. Maybe we can call it `ContractMetadataViewResolver`. How does that sound?

@sisyphusSmiling or @satyamakgec Can y'all check out the js changes to make sure those look okay?

austinkline:
> Thank you so much for adding this! My main suggestion is to choose a more specific name than `Resolver`. Maybe we can call it `ContractMetadataViewResolver`. How does that sound?

Open to any names folks generally agree on! I'll reach out to some other folks to see what they think as well, `ContractMetadataViewResolver` feels a bit long but agreed `Resolver` is a bit ambiguous

austinkline:
Spoke with some folks on the team and we like `ViewResolver` (h/t to @web3ashlee)

@joshuahannan does that work for you? If anyone else has preferences I'd love to hear them!

joshuahannan (approved):
It looks good to me, but we should try to get some more reviewers on it. Can you share in discord and I'll also try to get some more Flow team members to take a look at it

satyamakgec (approved):
LGTM

README.md
```diff
@@ -328,6 +328,35 @@ to your chosen fungible token for now. Then, use that public path for your royal
...
+
+Now that contract borrowing is released, you can also implement the [Resolver](./contracts/Resolver.cdc) interface on your contract
+and resolve views from there. As an example, you might want to allow your contract to resolve NFTCollectionData and NFTCollectionDisplay
+so that platforms do not need to find an NFT that belongs to your contract to get information about how to set up or show your collection.
+
+```cadence
+import ViewResolver from 0xf8d6e0586b0a20c7
+import MetadataViews from 0xf8d6e0586b0a20c7
+
+pub fun main(addr: Address, name: String): AnyStruct? {
```
sisyphusSmiling, line 341: Should this return `StoragePath?` instead of `AnyStruct?`? It would also be helpful to add this example to the `scripts/`directory.

sisyphusSmiling (approved):
Happy to see this! I think it's a great use case for the contract borrowing feature, and will be really helpful to have. Up to you if you'd like to act on the comment in this PR as it's more of a nitpick & additive.

README.md
austinkline: haha yeah this is `AnyStruct?` because i was playing around with what to make the test into, I get annoyed at changing the actual return format so I just leave it like this sometimes

Happy to change it, the main showcase is just how to resolve views in this way, what it returns isn't as big of a deal.

> It would also be helpful to add this example to the scripts/directory.

Noted!

joshuahannan:
I'm gonna merge this. Do we all think that it should just be deployed to the same account as `MetadataViews` and `NonFungibleToken`? In hindsight, we should have deployed `MetadataViews` to a different account, so I just want to check if we maybe think this would be better in a different account too

PR #168, opened 2023-04-09
remove file path syntax for imports

austinkline:
Closes: #???

## Description

Removes the file path imports since new versions of the flow cli don't need it. If we do this, you can use submodules to pull in standard repos so you don't have to copy other contracts anymore!

______

For contributor use:


bluesign:
shouldn't be something like: `import "NonFungibleToken"` instead of `import NonFungibleToken from "NonFungibleToken"`

austinkline:
> shouldn't be something like: `import "NonFungibleToken"` instead of `import NonFungibleToken from "NonFungibleToken"`

You're right! Tried that just now and it works as well. I could have sworn when I tried it a week back, my imports didn't resolve. I will amend this to use:

```cadence
import "MetadataViews"
```

instead of 

```cadence
import MetadataViews from "MetadataViews"
```

austinkline:
Looks like changing syntax made js tests break. Does flow-js-testing support that yet?

joshuahannan:
Yeah, we weren't ready to update to the new syntax because it wasn't compatible with our existing tests and repo organization yet, but this might work now. @sisyphusSmiling or @satyamakgec can you take a look at the js tests and see what might be going wrong?

austinkline:
> Yeah, we weren't ready to update to the new syntax because it wasn't compatible with our existing tests and repo organization yet, but this might work now.

I can always revert the last commit of this PR and we can tackle the final bit of it once support is there, I'm not blocked on this so whatever folks on your end would rather

Issue #106, opened 2022-07-12
Metadata Views: Full License View


joshuahannan:
### Issue To Be Solved
Dapper Labs is working with governments to define a standard for how NFT projects can define licensing rights for NFTs that define permissions for what can be done with them and their metadata by the owner. 
The current License view does not sufficiently cover the needs for a full license view, so we would like to define a new one that projects can start using. 

### Suggest A Solution
* Decide if we want to define a new view, or modify the old one. (probably need to define a new one)
* Decide which fields are needed
Options:
    * plain text description
    * ipfs hash
    * file
    * multiple licenses?

We'll post and discuss more in this issue

joshuahannan:
@psiemens @EricLin2004 Would you be able to elaborate a bit more on what is required for this? I want to make sure this is pretty clear

joshuahannan:
Here are two documents that our legal team has put together for how we envision this. I'll try to respond again later this week with some ideas for how I might want to implement this but any ideas or feedback in the meantime is appreciated. :)

[NFT License Grouping Model.pdf](https://github.com/onflow/flow-nft/files/9133655/NFT.License.Grouping.Model.pdf)
[Overview of NFT License Grouping Model.pdf](https://github.com/onflow/flow-nft/files/9133656/Overview.of.NFT.License.Grouping.Model.pdf)

joshuahannan:
![Overview of NFT License Grouping Model](https://user-images.githubusercontent.com/7407828/179599781-ca2f92e8-1c3b-48a2-8a21-90eec153939c.png)

bluesign:
This is very good idea, as long as Flow Team ( blockchain ) does not enforce this, just for user information and user trust, I think it will be beneficial to ecosystem. 

But from the look of the chart seems grouping is not so well. 

- Why `Voting Rights` and `Content and Experiences` mutually exclusive?

If we exclude the full ownership for a second: 

- There are 5 Groups ( 3 Colors ) with first 6 license items out of 10 are same. 

We can summarize with 2 questions: 
- Do you want commercial license? ( no / limited / unlimited ) 
- Do you want additional utility? ( Voting / Content / Skip )

I think very poorly designed tbh. Maybe even multiple badges can cover better.  ( e.g.  base & commercial & voting )

Even technically those rights can be interfaces, and can be combined later. Or you can maybe upgrade your NFT later for another license. 

- "subject to potential approval of content creator" I think here can be a little push on the content provider side.

[...]

PR #172, opened 2023-05-31
first stab at adding nft license view 

bjartek:
I added methods for each of the 12 combinations. 

Maybe we do not need to store this as a map anymore as now you cannot change the view after you have created it. Not really sure how strong the validation rules should be here. 

## Description

______

For contributor use:

### DIFF
--- a/contracts/MetadataViews.cdc
+++ b/contracts/MetadataViews.cdc
@@ -288,7 +288,7 @@ pub contract MetadataViews {
         }
         return nil
     }
-    
+
     /// View that defines the composable royalty standard that gives marketplaces a 
     /// unified interface to support NFT royalties.
     ///
@@ -395,8 +395,8 @@ pub contract MetadataViews {
         pub let mediaType: String
 
         init(file: AnyStruct{File}, mediaType: String) {
-          self.file=file
-          self.mediaType=mediaType
+            self.file=file
+            self.mediaType=mediaType
         }
     }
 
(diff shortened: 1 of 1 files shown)


joshuahannan (commented):
Thanks for taking a stab! I was a little confused about how you implemented it at first, but after thinking about it, I think it makes sense and allows us to add more combinations in the future if needed. I'm wondering if it also would be nice to also include a description like we did in the proposal that I made in [the other issue](https://github.com/onflow/flow-nft/issues/106). 

And also similar to that proposal, we want to include a `Media` for each right to potentially link to an off-chain description of the rights or this "badge" that is mentioned in the proposal.

I'm happy to make those changes and add some tests, though I'm not sure the best way to do it since this is in your fork. Can I just copy your code to my branch?

bjartek:
> Thanks for taking a stab! I was a little confused about how you implemented it at first, but after thinking about it, I think it makes sense and allows us to add more combinations in the future if needed. I'm wondering if it also would be nice to also include a description like we did in the proposal that I made in [the other issue](https://github.com/onflow/flow-nft/issues/106).
> 
> And also similar to that proposal, we want to include a `Media` for each right to potentially link to an off-chain description of the rights or this "badge" that is mentioned in the proposal.
> 
> I'm happy to make those changes and add some tests, though I'm not sure the best way to do it since this is in your fork. Can I just copy your code to my branch?

Feel free to copy

joshuahannan:
I opened a new PR with my suggestions: https://github.com/onflow/flow-nft/pull/173

I'll close this for now in favor of that one. Please leave feedback when you get the chance! Thank you. :)

PR #187, opened 2023-09-06
Fix missing auth in provider type

bjartek:


joshuahannan (approved):
Great catch! Thank you. 😄

PR #192, opened 2023-10-19
Update Cadence tests to latest API features

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/216

## Description

Updates the tests to be compatible with the upcoming version of the Cadence testing framework.

______

For contributor use:

- [x] Code follows the [standards mentioned here](https://github.com/onflow/flow-ft/blob/master/CONTRIBUTING.md#styleguides).
- [x] Updated relevant documentation 


scripts/borrow_nft.cdc
```diff
@@ -1,7 +1,7 @@
 // This script borrows an NFT from a collection
 
-import NonFungibleToken from "NonFungibleToken"
-import ExampleNFT from "ExampleNFT"
+import "NonFungibleToken"
+import "ExampleNFT"
```
joshuahannan, line 4: I'm confused why changing these to this kind of import doesn't break the go tests?

joshuahannan (approved):

scripts/borrow_nft.cdc
m-Peter: Maybe the specific script is not part of a test case? :thinking:

joshuahannan: I meant for all of them. The go tests expect everything to be in the form of import NonFungibleToken from "NonFungibleToken" and it replaces the placeholder with an actual address, so in this case it would just replace it with import 0x.... Is that valid if there is only one contract at that address?

It is used in the tests here: https://github.com/onflow/flow-nft/blob/master/lib/go/test/nft_test.go#L56

m-Peter: Hey @joshuahannan, good point.

I did check how that works, and the end result of the script looks like:
```cadence
Script:  // This script borrows an NFT from a collection

import 0x01cf0e2f2f715450
import 0xe03daebed8ca0615

pub fun main(address: Address, id: UInt64) {
    let account = getAccount(address)

    let collectionRef = account
        .getCapability(ExampleNFT.CollectionPublicPath)
        .borrow<&{NonFungibleToken.CollectionPublic}>()
        ?? panic("Could not borrow capability from public collection")

    // Borrow a reference to a specific NFT in the collection
    let _ = collectionRef.borrowNFT(id: id)
}
```

An import of the form `import 0xe03daebed8ca0615` means import every contract from that address. However, you can't have the same import for an address twice:

```cadence
import 0xe03daebed8ca0615
import 0xe03daebed8ca0615
```

This would throw an error.

joshuahannan: okay, that makes sense. So fine for now, but we might need to update it in the future if we decide to have contracts at the same address

PR #196, opened 2023-12-01
added updated event

bjartek:
Added updated event and a method to emit it.

### DIFF
--- a/contracts/NonFungibleToken-v2.cdc
+++ b/contracts/NonFungibleToken-v2.cdc
@@ -206,7 +206,7 @@ access(all) contract NonFungibleToken {
         access(all) view fun borrowNFTSafe(id: UInt64): &{NonFungibleToken.NFT}? {
             post {
                 (result == nil) || (result?.getID() == id): 
-                "Cannot borrow NFT reference: The ID of the returned reference does not match the ID that was specified"
+                    "Cannot borrow NFT reference: The ID of the returned reference does not match the ID that was specified"
             }
             return nil
         }
--- a/contracts/NonFungibleToken-v2.cdc
+++ b/contracts/NonFungibleToken-v2.cdc
@@ -51,6 +51,17 @@ access(all) contract NonFungibleToken {
     // An entitlement for allowing the withdrawal of tokens from a Vault
     access(all) entitlement Withdrawable
 
+    /// Event that is emitted when a token is updated,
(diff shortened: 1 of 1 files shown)


contracts/NonFungibleToken-v2.cdc
```diff
@@ -195,7 +206,7 @@ access(all) contract NonFungibleToken {
         access(all) view fun borrowNFTSafe(id: UInt64): &{NonFungibleToken.NFT}? {
             post {
                 (result == nil) || (result?.getID() == id): 
-                    "Cannot borrow NFT reference: The ID of the returned reference does not match the ID that was specified"
+                "Cannot borrow NFT reference: The ID of the returned reference does not match the ID that was specified"
```
bjartek, line 209:
```suggestion
                    "Cannot borrow NFT reference: The ID of the returned reference does not match the ID that was specified"
```

contracts/NonFungibleToken-v2.cdc
```diff
@@ -51,6 +51,17 @@ access(all) contract NonFungibleToken {
     // An entitlement for allowing the withdrawal of tokens from a Vault
     access(all) entitlement Withdrawable
 
+    /// Event that is emitted when a token is updated,
+    ///
+    access(all) event Updated(id: UInt64, uuid: UInt64, owner: Address, type:String)
```
joshuahannan, line 56: Not sure if this is actually possible, but when you commented about it. I assumed that the event would include some sort of information about what was updated. Like an `AnyStruct`, a string description of what was updated, or a custom struct that we can define that includes multiple fields that elaborate on the update. Do you think something like that would be useful? I feel like with the way that the event currently is, indexers wouldn't get any useful information from this.

bjartek: What we do in the indexer for Doodles is that we just reevaluate all the views when this happends. We can always add a context map though.

joshuahannan: like you just query that owners account and check all the views for the doodle that was updated? I guess that works. idk if it would even be possible in a generic way to make that information accessible in an event

joshuahannan: We can discuss in the meeting next week

joshuahannan (approved):

Issue #129, opened 2022-09-20
A new Metadata view to interop with NFT collections that store NFT in a offchain json file


bjartek:
DimentionX and Rarible both store nfts in an offchain json file that follows the openSea standard. I think this should be formalized into a Struct. 

I vaguely remeber that we have talked about this before @psiemens

austinkline:
That was part of the the NFTView is for, would the goal be to translate the format from the OS struct to the NFTView?

spacepluk:
I'm very happy to see this here :D

psiemens:
When we talked about this before, the idea was to have a simple view that returns a URL to a JSON file.

I think it'd basically just be a port of the ERC 721 / 1155 `tokenURI` function: https://docs.opensea.io/docs/metadata-standards#implementing-token-uri

As @bjartek mentioned, there are some existing projects that store metadata this way. I know that many of us are champions of on-chain data (myself included), but I still think its valid for off-chain metadata like this to be given a home inside a view.

joshuahannan:
Anybody have any ideas for the implementation for this?

bjartek:
A naive implementation would just be to have a struct that has a File, but then you would lose mediaType information, does anybody use anything but json for this?

Another option is to use Medias so that you can set the mediaType to what you want like application/json+opensea?

On ETH/opensea this is called `tokenURI` but I am not sure if that name is appropriate here. What about OffchainMetadataURI?

so what about
```
pub stuct OffchainMetadataURI{
   pub let media:Media
   
   init(_ media:Media) {
    self.media=media
  }

  pub fun uri() : String {
    return self.media.uri()
  }
}
```

austinkline:
> Anybody have any ideas for the implementation for this?

I would strongly suggest that we lay out an actual data format that is expected. This problem to me is framed with the following question:

"What is the purpose of exposing off-chain data, and who is expected to use it?"

My answer to that question is that exposing data this way is almost exclusively for third-party platforms. Products which store data off-chain such as Solarpups already know how their data is formatted for their own system to work. Guiding products to a specific unified (but flexible) format will be extremely important.

I think we have two options.
1. As @spacepluk has asked for in the past, we endorse vending all contents of the `NFTView` struct from this external URL
2. We encourage only a subset such as Traits and Display.

[...]

PR #203, opened 2024-03-12
Add EVMBridgedMetadata & URI to MetadataViews

sisyphusSmiling:
Related: #129 

## Description

With EVM on Flow coming in the near future, we'll need to reconcile differences between EVM & Cadence NFT metadata handling expectations and standards.

Motivated by ongoing work while iterating on the VM bridge, this PR puts forth two metadata views for consideration. 

The first, `URI`, is a generic struct intended to store a string pointer to some metadata. In many cases, this may just be a JSON blob at a URL or an IPFS pointer. In other cases, this may actually be NFT metadata as a JSON blob itself as in URL-encoded URIs. In the case of ERC721s bridged from EVM, we likely won't have enough context to determine the type of URI this is, thus the generic nature of this view.

Here is the `URI` interface:

```cadence
pub struct URI: MetadataViews.File {
    /// The base URI prefix, if any. Not needed for all URIs, but helpful
    /// for some use cases For example, updating a whole NFT collection's
    /// image host easily
    pub let baseURI: String?
    /// The URI string value
    /// NOTE: this is set on init as a concatenation of the baseURI and the
    /// value if baseURI != nil
    access(self) let value: String

    pub view fun uri(): String
        
}
```

The second, `EVMBridgedMetadata`, contains three fields 

```cadence
pub struct EVMBridgedMetadata {
    pub let name: String
    pub let symbol: String

    pub let uri: {MetadataViews.File}
}
```

The fields `name` and `symbol` are included in several ERC standards, notably `ERC721` and `ERC20`. The last field, `uri`, allows Cadence projects to define how they would like their token or contract's metadata to be transmitted to an EVM counterpart. In the case of Cadence-native projects, this field could be any `File` implementing struct, while in the case of bridged assets (e.g. ERC721) the `uri` value would likely just be a `URI` struct.

Without getting into the specifics of bridging assets which is out of scope for this PR, it would be very helpful to get feedback on the views as their presented, if we feel they're worth adding in the MetadataViews contract, and if the community feels anything is missing from these very barebones implementations.

______

For contributor use:


sisyphusSmiling:
[Relevant discussion in discord](https://discord.com/channels/613813861610684416/1217248085319749663)

joshuahannan (approved):
This makes sense to me. Would you want to deploy this upgrade to `MetadataViews` before the Crescendo upgrade so people can start adding it immediately, or do we want to deploy this as part of the Crescendo upgrade? If the latter is true, we should base this on the `standard-v2` branch

sisyphusSmiling:
Updating to note I added a `baseURI: String?` field to `URI` based on [feedback](https://github.com/onflow/flow-evm-bridge/pull/14#discussion_r1526761024) implementing the view in bridge contracts.

@joshuahannan I think deploying the update before Crescendo makes the most sense to allow projects to adopt the view as they see fit before bridging is enabled. There was constructive discussion in the linked discord thread, but I'd hoped for more community feedback before merging this view. I'll raise this PR again in discord to ask folks to share feedback.

joshuahannan (approved):
We want to do this upgrade before Cadence 1.0, right? After we merge this, can you make a PR updating the V2 standards to include this and also add an implementation for it in the V2 ExampleNFT? I can do the upgrades on testnet and mainnet

sisyphusSmiling:
Yeah, we'd want to include this before 1.0. I can follow up with a PR including an implementation.

joshuahannan:
oh yeah, and make sure you run `make test` to run the tests and generate the assets

PR #205, opened 2024-03-21
Make ExampleNFT compatible for contract update

SupunS:
Related to https://github.com/onflow/cadence/issues/3098#issuecomment-2007977094.

I don't know much about the use-cases, so I only updated it to be "semantically valid". So it may not be the best way to update.

## Description

Made the contract compatible for updating by reverting the following changes:
- Removed the two new fields `storagePath` and `publicPath`.
- Changed back the type of `ownedNFTs` to `@{UInt64: NonFungibleToken.NFT}`
- Brought back the `ExampleNFTCollectionPublic` interface. 
- Brought back the conformance of `Collection` to `ExampleNFTCollectionPublic`.

______

For contributor use:


contracts/ExampleNFT.cdc
```diff
@@ -117,26 +117,23 @@ access(all) contract ExampleNFT: NonFungibleToken {
...
+    access(all) resource interface ExampleNFTCollectionPublic {}
+
+    access(all) resource Collection: NonFungibleToken.Collection, ExampleNFTCollectionPublic {
         /// dictionary of NFT conforming tokens
         /// NFT is a resource type with an `UInt64` ID field
-        access(contract) var ownedNFTs: @{UInt64: ExampleNFT.NFT}
-
-        access(all) var storagePath: StoragePath
-        access(all) var publicPath: PublicPath
+        access(contract) var ownedNFTs: @{UInt64: {NonFungibleToken.NFT}}
```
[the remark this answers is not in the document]
SupunS: Yeah I understand the concerns.

The reason for not allowing changing the field is, if the type was changed in the contract, then all of a sudden the existing data stored no longer match the type of this field. So in order make it compatible, we would need to also migrate the values to the changed type, by running a migration. However, this is going to be very complex because:
1) We would have to keep track of all the field types changes, and then during the migration, we would need to backtrack and find the type of the field, and then change the type of value accordingly.
2) Some changes to types may not be compatible, but we can only know that for sure during the migration of the value. For e.g: given the type of the map's values were previously `NonFungibleToken.NFT`, it is theoretically possible to have any `NonFungibleToken.NFT` in the map, and not just `ExampleNFT.NFT` (maybe not in this particular example, but possible in other similar cases). i.e: This is "narrowing" the type, and narrowing isn't always guaranteed to be succeed. Where as broadening the type (changing from `ExampleNFT.NFT` to `NonFungibleToken.NFT`) would. So given that the contract is already staged by the time we run the migrations, if we run into such incompatibilities, those data would be forever corrupted. So that's why the contract update validator only allows changes that are 100% guaranteed to be valid.

Again, I get this may not be ideal, maybe we could do some brainstorming and try to find a middle ground.

contracts/ExampleNFT.cdc
```diff
@@ -304,8 +301,11 @@ access(all) contract ExampleNFT: NonFungibleToken {
 
         // Create a Collection resource and save it to storage
         let collection <- create Collection()
-        let defaultStoragePath = collection.storagePath
-        let defaultPublicPath = collection.publicPath
+
+        let identifier = "cadenceExampleNFTCollection"
```
nvdtf, line 305: For compatibility with previous paths
```suggestion
        let identifier = "exampleNFTCollection"
```
L232 and L233 need to be updated too
For reference, here's the previous version: https://github.com/onflow/flow-nft/blob/4c0c684baed98e2ffd7a8f9d27973c1db5dd1239/contracts/ExampleNFT.cdc#L390

contracts/ExampleNFT.cdc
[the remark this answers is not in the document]
joshuahannan: okay that is fine for now. I mostly want it there as an example for people who are creating new collections, but I guess we can just change it after the cadence 1.0 upgrade is complete

contracts/ExampleNFT.cdc
joshuahannan: @SupunS Were you going to make this change or should I just do it?

SupunS: Added in https://github.com/onflow/flow-nft/pull/205/commits/692b3dea2d0051b35516c6754ac7cc05de1d434b

SupunS: oh, just realize there are many places that needs updating. @joshuahannan might be better if you can update it. I could break things unknowingly.

joshuahannan: okay. I'll pull down your branch and make the changes

joshuahannan (approved):

contracts/ExampleNFT.cdc
```diff
@@ -16,13 +16,17 @@ import "MetadataViews"
 
 access(all) contract ExampleNFT: NonFungibleToken {
 
+    /// Standard Paths
+    access(all) let CollectionStoragePath: StoragePath
+    access(all) let CollectionPublicPath: PublicPath
```
SupunS, line 21: Adding new fields would get rejected during the contract update

joshuahannan: These aren't new. They existed in the old version, so since we're expecting people to follow this as an example for how to update their contracts, keeping them makes sense

PR #207, opened 2024-03-26
Remove INFT

### LINKED ISSUE #206: [standard v2] Remove `INFT` interface from `NonFungibleToken`
### Issue To Be Solved
I guess the reason for keeping that was to make the `NonFungibleToken` backward compatible, and to make the life easier for the migrations.

Now that the migrations and contract update validator support changing the type from  `NonFungibleToken.INFT` to `NonFungibleToken.NFT` (added in https://github.com/onflow/flow-go/pull/5564), we could remove the old `INFT` interface, and ask developers to start using the new `NFT`.

In-fact now it is also unsafe from migrations point of view, to have both the `INFT` and the `NFT` around, because it could theoretically cause "type confusion" for certain migrated values. So removing the `INFT` interface, and requiring people to start using `NFT` instead seems the best option.

turbolent:
Closes #206

### DIFF
--- a/contracts/NonFungibleToken.cdc
+++ b/contracts/NonFungibleToken.cdc
@@ -85,9 +85,6 @@ access(all) contract interface NonFungibleToken: ViewResolver {
     ///
     access(all) event Deposited(type: String, id: UInt64, uuid: UInt64, to: Address?, collectionUUID: UInt64)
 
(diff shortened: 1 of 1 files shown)


joshuahannan:
closing in favor of https://github.com/onflow/flow-nft/pull/205

PR #210, opened 2024-04-03
Reorganize NFT transactions and scripts

joshuahannan:
## Description
* Renames tests to end with `_test.cdc`
* Reorganize scripts into folders within the `transactions/` directory and keep scripts that are purely for tests in the `tests/` directory
* Fix a few metadata tests

I'm planning on going through the docs site and cadence-lang.org to make sure all the links to these moved transactions and scripts are fixed

I mostly want to know if we think it is a bad idea to change all these paths since it might break links or dependencies in the ecosystem that link to these scripts. It would be nice to standardize them though so we can have a unified structure among all our repos


joshuahannan:
I only found one link across the `onflow/docs` repo and `cadence-lang.org` repo that needed to be updated, so I make a PR for that: https://github.com/onflow/docs/pull/712. It makes me feel better about the other updated links

PR #126, opened 2022-09-09
FEATURE: V2 NonFungibleToken Standard

joshuahannan:
Adds the v2 fungible token standard, as described in this flip: https://github.com/onflow/flips/pull/56

A few highlights:

* Turns NonFungibleToken into a contract instead of a contract interface
* Introduces a new contract interface (`NonFungibleTokenInterface`) with event specifications, functions to get the NFT and collection types that the contract implements, and functions to get important metadata views for the types in the contract. This will help external parties learn about which types a contract implements since contracts can implement multiple collection and NFT types now.
* Adds a Transferable interface to include a transfer function
* Adds getAcceptedTypes() to the Receiver interface so a receiver interface can indicate which NFT types it accepts.
* Moves standard paths inside the collection object definition because they are specific to the actual collection instead of the contract now.
* Adds type parameters to the all the events since contracts can have multiple nft types defined
* Adds a type parameter to the `createEmptyCollection` method

The contract currently has a couple errors that should be able to be resolved easily, such as it not working properly with the `createEmptyCollectionFunction` field of the `NFTCollectionData` metadata view, and the `ownedNFTs` field not conforming to the resource interface's field, but there are no other errors.
I am looking for feedback on the initial proposal. Tests will come later
There are probably more things that I am missing, but most of the details are in the forum post, so that is another good resource.

PR #211, opened 2024-04-11
Add NFT Storage Requirement to Collection and `forEachID()`

joshuahannan:
## Description

[FLIP](https://github.com/onflow/flips/pull/258)

Based on [this discussion](https://discord.com/channels/613813861610684416/621847426201944074/1227401046926692443) in discord and [this comment](https://github.com/onflow/flow-nft/pull/126#issuecomment-2048225823) on the NFT standard PR, we have come to the conclusion that it probably makes sense to enforce how NFTs are stored for standardization and in this specific case, to allow projects to easily iterate through IDs in a collection that has many stored NFTs.

Enforcing the `ownedNFTs` field does not restrict projects who want to store their NFTs in unique ways because they can just ignore the field if they want, so having it for the 99% of projects who do and will follow it is very useful.

______

For contributor use:

- [x] Targeted PR against `standard-v2` branch
- [x] Updated relevant documentation 


contracts/ExampleNFT.cdc
```diff
@@ -150,11 +150,11 @@ access(all) contract ExampleNFT: NonFungibleToken {
...
-           if type == Type<@ExampleNFT.NFT>() {
-            return true
-           } else {
-            return false
-           }
+            if type == Type<@ExampleNFT.NFT>() {
+                return true
+            } else {
+                return false
+            }
```
turbolent, line 157:
```suggestion
            return type == Type<@ExampleNFT.NFT>()
```

contracts/NonFungibleToken.cdc
```diff
@@ -194,7 +197,16 @@ access(all) contract interface NonFungibleToken: ViewResolver {
...
         /// @return An integer indicating the size of the collection
-        access(all) view fun getLength(): Int
+        access(all) view fun getLength(): Int {
+            return self.ownedNFTs.length
+        }
+
+        /// Returns an iterator that allows callers to iterate
+        /// through the list of owned NFT IDs in a collection
+        /// without having to load the entire list first
+        access(all) fun forEachID(_ f: fun (UInt64): Bool): Void {
```
turbolent, line 207:
```suggestion
        access(all) fun forEachID(_ f: fun (UInt64): Bool) {
```

transactions/scripts/iterate_ids.cdc
```diff
@@ -0,0 +1,31 @@
...
+
+access(all) fun main(ownerAddress: Address, limit: Int): Int {
+
+    let response: [&{NonFungibleToken.NFT}] = []
+
+    let account = getAuthAccount<auth(BorrowValue) &Account>(ownerAddress)
+
+    account.storage.forEachStored(fun (path: StoragePath, type: Type): Bool {
+
+        if (!type.isSubtype(of: Type<@{NonFungibleToken.Collection}>())) {
```
turbolent, line 11:
```suggestion
        if !type.isSubtype(of: Type<@{NonFungibleToken.Collection}>()) {
```

transactions/scripts/iterate_ids.cdc
```diff
@@ -0,0 +1,31 @@
...
+
+            return true
+        }
+
+        let storageCollection = account.storage.borrow<&{NonFungibleToken.Collection}>(from: path)!
```
turbolent, line 16: Checking the type isn't really necessary, given the borrow checks the type anyway.

```suggestion
        if let storageCollection = account.storage.borrow<&{NonFungibleToken.Collection}>(from: path) {
            // ...
        } else {
            return false
        }
```

contracts/NonFungibleToken.cdc
joshuahannan: I'm going to keep the `Void` here because I think it makes it more clear

contracts/ExampleNFT.cdc
```diff
@@ -134,7 +134,7 @@ access(all) contract ExampleNFT: NonFungibleToken {
     access(all) resource Collection: NonFungibleToken.Collection, ExampleNFTCollectionPublic {
         /// dictionary of NFT conforming tokens
         /// NFT is a resource type with an `UInt64` ID field
-        access(contract) var ownedNFTs: @{UInt64: {NonFungibleToken.NFT}}
+        access(all) var ownedNFTs: @{UInt64: {NonFungibleToken.NFT}}
```
sisyphusSmiling, line 137: Might be worth adding a note in this comment regarding access control since the interface is `access(contract)`. I didn't realize subtypes could define less restrictive access

contracts/NonFungibleToken.cdc
```diff
@@ -194,7 +197,16 @@ access(all) contract interface NonFungibleToken: ViewResolver {
...
         /// Gets the amount of NFTs stored in the collection
         /// @return An integer indicating the size of the collection
-        access(all) view fun getLength(): Int
+        access(all) view fun getLength(): Int {
+            return self.ownedNFTs.length
+        }
+
+        /// Returns an iterator that allows callers to iterate
+        /// through the list of owned NFT IDs in a collection
+        /// without having to load the entire list first
```
sisyphusSmiling, line 206: Same suggestion as committed to [the FLIP](https://github.com/onflow/flips/pull/258)

```suggestion
        /// Allows a given function to iterate through the list
        /// of owned NFT IDs in a collection without first
        /// having to load the entire list into memory
```

contracts/NonFungibleToken.cdc
```diff
@@ -194,7 +197,16 @@ access(all) contract interface NonFungibleToken: ViewResolver {
...
+        access(all) fun forEachID(_ f: fun (UInt64): Bool): Void {
+            self.ownedNFTs.forEachKey(f)
```
sisyphusSmiling, line 208: If I define a function that removes the NFT from the mapping and either destroys or transfers it, would that be possible? Maybe we add a post condition on length to the method

cody-evaluate: @sisyphusSmiling i dont fully understand, by post condition do you mean an assert inside of the `forEachID` method?

sisyphusSmiling: That all makes sense, thanks for clarifying. I was initially spooked to see a non-view method running arbitrary code on `self.ownedNFTs` within interface scope, but I can see how a given method would still need entitled access on the underlying collection or `ownedNFTs` mapping. The post-condition was misguided, so no worries there.

sisyphusSmiling (approved):

Issue #232, opened 2024-09-09
Update NFT Error messages to be more descriptive


joshuahannan:
Works towards https://github.com/onflow/docs/issues/795

### Issue To Be Solved
Error messages for panics in transactions and contracts are not very helpful to end-users.
They should be updated to be more descriptive

### Suggest A Solution
Go through all the standard contracts, transactions, and scripts to update them with good error messages. Error messages could contain:
* Contract name and function name if coming from a contract
* Description of literal error
* Description of what high-level reason might be causing the error
* Suggestion for fixing it if relevant

PR #233, opened 2024-09-09
Adds more descriptive and helpful error messages to the NFT contracts and txs

joshuahannan:
Closes: #232

## Description

* Updates all error messages to include more information and relevant data about the error if possible
* Updates CLI commands for C1.0

______

For contributor use:


contracts/ExampleNFT.cdc
```diff
@@ -133,7 +133,7 @@ access(all) contract ExampleNFT: NonFungibleToken {
                             resourceType: nil,
                             viewType: Type<MetadataViews.EVMBridgedMetadata>()
                         ) as! MetadataViews.EVMBridgedMetadata?
-                        ?? panic("Could not resolve contract-level EVMBridgedMetadata")
+                        ?? panic("Could not resolve contract-level EVMBridgedMetadata.")
```
bjartek, line 136: Imho resolving a view should never
Panic. Return nil.

sisyphusSmiling (approved):
Really like the detailed error message, thanks for fixing these up! My main suggestion is not a blocker, but it would be great to see consistent formatting e.g. some messages wrap values in `()` or `<>` and others prefix values with `=`. In reality, this probably more of a preference/nitpick

contracts/ExampleNFT.cdc
joshuahannan: Agreed! I'll remove that. Good catch!

joshuahannan:
@sisyphusSmiling Good point! I'll look through and make sure they are consistent. The only time I use `<>` is when I am dealing with types because that is usually what they are contained in in the code, so I thought it would make sense to be consistent with that. But agreed about using `=` vs `()`. I'll find a consistent format.

Issue #242, opened 2025-02-06
Add CrossVM views


sisyphusSmiling:
Related: https://github.com/onflow/flips/issues/318

## Description

Add the two views proposed in [FLIP 318](https://github.com/onflow/flips/pull/319) - `EVMPointer` and `EVMBytesMetadata`.

Since adding these views to the standard `MetadataViews` would result in a circular dependency (`EVM > MetadataViews > FlowToken > EVM`), these views will need to be added to an alternate contract. While unfortunate and might normally result in reduced adoption, these views are ultimately only used by the bridge when moving NFTs between Cadence & EVM. Since the implementation of these views are a hard requirement for metadata to be passed into EVM, they will be leveraged in any documentation, guides and examples demonstrating dual VM deployment and can be assumed developers would see their documented use in the process of writing their contracts.

PR #243, opened 2025-02-06
Add CrossVM views

sisyphusSmiling:
Closes: #242
Related: https://github.com/onflow/flow-evm-bridge/issues/164

## Description

- Adds views as proposed in https://github.com/onflow/flips/issues/318 in a new contract `CrossVMMetadataViews`
- Updates flow.json

______

For contributor use:


joshuahannan (approved):
Would you be able to add examples of these to `ExampleNFT` and add some tests?

sisyphusSmiling:
@joshuahannan I've added implementation into ExampleNFT and tests. lmk if you think those are sufficient. 

Cadence tests are passing locally, but CI is failing on the go tests. I think I'll have to get the EVM contract into the go test env as they're failing locally as well.

We can also add more use case specific examples for Cadence- and EVM-native NFT collections once the bridge changes are complete.

sisyphusSmiling:
Looks like CI is failing due to the changes in the ExampleNFT. When emulator is bootstrapped, it deploys `ExampleNFT` ([ref](https://github.com/onflow/flow-emulator/blob/88c5a70d1580b2db59a21a8f516eda29e2421648/emulator/contracts.go#L39)) and since the contract in this repo now requires two new imports, the bootstrap fails. Not sure how to get around this @joshuahannan as this is a bit circular

joshuahannan (approved):
Yeah, this happens sometimes. We'll just need to tag a prerelease based on this branch and use that in flow-go, then when the emulator is updated, we can merge this.

joshuahannan:
Another option would be to add a separate contract getter to the `contracts` package to get the new contract, and use that to test. Probably should do that actually so we can still make sure the go tests are passing

Issue #250, opened 2025-07-24
Add NFT function to resolve a view from a type identifier


joshuahannan:
### Issue To Be Solved
Currently transactions have to have a lot of boilerplate code to get specific metadata views from a type identifier
It would be good if there was a utility function for this.

### Suggest A Solution
Add a function `view fun resolveViewFromTypeIdentifier(identifier: String, view: Type): AnyStruct?` that takes a type identifier String and view to resolve, then gets the type from the identifier, borrow's the contract, and resolves the contract view that was requested. Returns `nil` if there is any error

PR #251, opened 2025-07-25
Adds `MetadataViews.resolveContractViewFromTypeIdentifier()` utility

joshuahannan:
Closes: https://github.com/onflow/flow-nft/issues/250

## Description

Adds a utility function to `MetadataViews` to resolve a view based on just a type identifier
Uses this function in the generic transactions to save code

______

For contributor use:


contracts/MetadataViews.cdc
```diff
@@ -14,6 +14,32 @@ import "ViewResolver"
...
+    /// Function to resolve a contract view based on a type identifier String
+    /// and view type. Borrows the contract as &{ViewResolver} and 
+    /// then calls resolveContractView for the specified type and view.
+    ///
+    /// @param resourceType: An optional resource type to return views for
+    /// @param viewType: The Type of the desired view.
+    /// @return A structure representing the requested view. If anything failed, returns nil
+    ///
+    access(all) fun resolveContractViewFromTypeIdentifier(
+        resourceTypeIdentifier: String,
```
sisyphusSmiling, line 26: I don't feel super strongly about this but it seems like for the sake of consistency at least the param should be `resourceType: Type` given `viewType: Type` and `resolveContractView()` param types. If so, the method needs renaming.

If `*Identifier` makes more sense based on how this will be called, then we should update `@param` to match the parameter name and type.

sisyphusSmiling (approved):
Nice addition! This will save quite a bit of boilerplate.

contracts/MetadataViews.cdc
joshuahannan: oh yeah, I just forgot to update the comment. I think it is better to do it as the identifier since it saves more code

